PR #19, opened 2024-10-02
Improvement: Refactor Flask Backend for Maintainability (Edited by maintainer)

VarunParab:
Changes Made:

- Added a VideoProcessor class to encapsulate video download and processing logic.
- Moved methods like download_youtube_video() (renamed to download_media()) and process_video_with_cpp() into VideoProcessor.
- Encapsulated file path sanitization and cleanup operations into the Utils class.
- Implemented static methods in the class, this provides a clear and concise way to handle utility operations without requiring an instance of the class.

### DIFF
--- a/app.py
+++ b/app.py
@@ -14,6 +14,7 @@ How it works:
 2) Send a processing request to the `MediaProcessor` C++ binary, which uses DeepFilterNet for fast filtering
 3) Serve the processed video on the frontend
 
+Refactored to improve maintainability and readability.
 """
 
 app = Flask(__name__)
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -14,6 +14,7 @@
 2) Send a processing request to the `MediaProcessor` C++ binary, which uses DeepFilterNet for fast filtering
 3) Serve the processed video on the frontend
 
+Refactored to improve maintainability and readability.
```
omeryusufyagci, line 17: Let's remove this, as it will quickly turn into a large changelog, which is not the best place for it.

app.py
```diff
@@ -35,91 +36,105 @@
...
-                merged_ext = result.get('ext', 'mp4')
 
-        # Return the sanitized file path with for processing
-        video_file = os.path.join(app.config['UPLOAD_FOLDER'], sanitized_title + "." + merged_ext)
-        return os.path.abspath(video_file)
+class Utils:
+    """Utility class for common operations like file cleanup and sanitization."""
+
+    @staticmethod
+    def cleanup_old_files(base_filename):
```
omeryusufyagci, line 44: Let's rename this method to `remove_files_by_base`

app.py
```diff
@@ -35,91 +36,105 @@
...
+            # Return the sanitized file path for processing
+            video_file = os.path.join(app.config['UPLOAD_FOLDER'], sanitized_title + "." + merged_ext)
+            return os.path.abspath(video_file)
+
+        except Exception as e:
+            logging.error(f"Error downloading video: {e}")
+            return None
+
+    @staticmethod
+    def process_media_with_cpp(video_path):
```
omeryusufyagci, line 102: Let's rename this to `process_with_media_processor`. Since it's already a part of MediaHandler, no reason anymore to be explicit about we're processing media in the method name, but we should be explicit about the MediaProcessor.

app.py
```diff
@@ -35,91 +36,105 @@
...
+                logging.info(f"Removing old file: {path}")
+                os.remove(path)
+
+    @staticmethod
+    def sanitize_filename(filename):
+        """Replace non-alphanumerics (except periods and underscores) with underscores."""
+        return re.sub(r'[^a-zA-Z0-9._-]', '_', filename)
+
+
+class VideoProcessor:
```
omeryusufyagci, line 62: Let's rename this to MediaHandler for 2 reasons: we don't want to limit ourselves to videos yet, and media keeps it more generic, and we're not really processing media here, but rather handling this operation.

app.py
```diff
@@ -35,91 +36,105 @@
...
-        # Return the sanitized file path with for processing
-        video_file = os.path.join(app.config['UPLOAD_FOLDER'], sanitized_title + "." + merged_ext)
-        return os.path.abspath(video_file)
+class Utils:
+    """Utility class for common operations like file cleanup and sanitization."""
+
+    @staticmethod
+    def cleanup_old_files(base_filename):
+        """Removes any existing files with the same base name."""
+        file_paths = [
```
omeryusufyagci, line 46: for `file_paths` utilize the `base_filename` to avoid repetition; something like
```
base_path = os.path.join(app.config['UPLOAD_FOLDER'], base_filename)
file_paths = [
    base_path + '.webm',
    # and same for the isolated audio and processed video
]
```

app.py
omeryusufyagci, line 102: Let's rename this to `process_with_media_processor`. Since it's already a part of `MediaHandler`, no reason anymore to be explicit about we're processing media in the method name, but we should be explicit about the `MediaProcessor`.

app.py
```diff
@@ -35,91 +36,105 @@
...
-                processed_video_path = line.split(": ")[1].strip()
-                return processed_video_path
-
-        return None
-    except Exception as e:
-        logging.error(f"Error running C++ binary: {e}")
-        return None
 
 @app.route('/', methods=['GET', 'POST'])
 def index():
```
omeryusufyagci, line 126: Could you also add a generic URL validation? `urlparse` from `urllib.parse` should be a good start for us; better than nothing!

app.py
omeryusufyagci, line 17: Let's remove this, as it will quickly turn into a large changelog, which is not the best place for it.

app.py
omeryusufyagci, line 44: Let's rename this method to `remove_files_by_base`

app.py
omeryusufyagci, line 62: Let's rename this to `MediaHandler` for 2 reasons: we don't want to limit ourselves to videos yet, and media keeps it more generic, and we're not really processing media here, but rather handling this operation.
It's my bad for not catching this earlier, as I had proposed this name !

app.py
omeryusufyagci, line 46: for `file_paths` utilize the `base_filename` to avoid repetition; something like
```
base_path = os.path.join(app.config['UPLOAD_FOLDER'], base_filename)
file_paths = [
    base_path + '.webm',
    # and same for the isolated audio and processed video
]
```

app.py
```diff
@@ -35,91 +36,105 @@
 if not os.path.exists(app.config['UPLOAD_FOLDER']):
```
omeryusufyagci, line 36: We can add a util method `ensure_dir_exists` as this will likely reused

omeryusufyagci (changes requested):
Thanks for the great work so far, looks much better already! 

Could you please address these comments and let's take another look? Thanks!

VarunParab:
@omeryusufyagci I’ve made the changes you requested. Please have a look, and let me know if there’s anything else you’d like me to change.

omeryusufyagci:
@VarunParab looks good at a glance! but let me take a closer look over the weekend please. Thanks a lot!

omeryusufyagci (approved):
Looks good within the scope of this PR; approved.

Issue #5, opened 2024-09-29
Feature: Add Auto-Configuration for Thread Usage in `MediaProcessor`


omeryusufyagci:
Currently, the thread count uses a default value (see `MediaProcessor::AudioProcessor()`). This improvement aims to allow faster processing times for more capable hardware while ensuring stable operation on all target machines by determining an optimal number of threads.

**Objective:**
- Automatically determine the number of threads based on hardware capabilities using `std::thread::hardware_concurrency()`.
- Allow users to override this value in `config.json` for manual control if needed.
- Update the current implementation to use this dynamic value where applicable.

nikhiljangra264:
I would love to work on this, Could you assign it to me?

omeryusufyagci:
Hi @nikhiljangra264, thanks for your interest! I've assigned the issue to you.

nikhiljangra264:
Hello, I have made a PR regarding this. Please take a look and give me suggestions if any.

PR #21, opened 2024-10-03
Improvement: Add Auto-Configuration for Thread Usage in MediaProcessor #5

nikhiljangra264:
#5 

### Updates and Enhancements
- **Added `determineNumChunks()` function**  
  A new function in the `AudioProcessor` class is responsible for dynamically determining the number of chunks (threads) based on system hardware or configuration.
1. use number of chunks given in the config file if within range i.e. [1, total_threads].
1. else use total_threads - 2 if total_threads > 2 else use single thread
1. else use Deafult number of chunks if can't determine number of threads.

- **Updated `AudioProcessor` constructor**  
  The constructor calls the `determineNumChunks()` funtion to determine the dynamic number of chunks.

- **Modified `ConfigManager`**  
  The `ConfigManager` class now supports the configuration of `num_chunks` in the configuration file.

- **Updated the configuration file**  
  The configuration file has been updated to include the `num_chunks` setting, allowing users to adjust the number of chunks (threads) for processing.

### DIFF
--- a/MediaProcessor/src/AudioProcessor.cpp
+++ b/MediaProcessor/src/AudioProcessor.cpp
@@ -225,13 +225,13 @@ bool AudioProcessor::mergeChunks() {
     for (const auto& chunkPath : m_processedChunkPaths) {
         cmd.addFlag("-i", chunkPath.string());
     }
-  
+
     // If there's more then one processed chunk
     if (static_cast<int>(m_processedChunkPaths.size()) >= 2) {
         cmd.addFlag("-filter_complex", buildFilterComplex());
         cmd.addFlag("-map", "[outa]");
     }
-    
+
     cmd.addFlag("-c:a", "pcm_s16le");
     cmd.addFlag("-ar", "48000");
     cmd.addArgument(m_outputAudioPath.string());
--- a/MediaProcessor/src/AudioProcessor.cpp
+++ b/MediaProcessor/src/AudioProcessor.cpp
@@ -15,11 +15,13 @@ AudioProcessor::AudioProcessor(const std::string &inputVideoPath,
                                const std::string &outputAudioPath)
     : m_inputVideoPath(inputVideoPath),
       m_outputAudioPath(outputAudioPath),
-      m_numChunks(DEFAULT_NUM_CHUNKS),
       m_overlapDuration(DEFAULT_OVERLAP_DURATION) {
     m_outputDir = std::filesystem::path(outputAudioPath).parent_path().string();
     m_chunksDir = m_outputDir + "/chunks";
     m_processedChunksDir = m_outputDir + "/processed_chunks";
+
+    m_numChunks = ConfigManager::getInstance().getOptimalThreadCount();
+    std::cout << "INFO: using " << m_numChunks << " threads." << std::endl;
 
     /*
      * TODO: 
@@ -171,19 +173,15 @@ bool AudioProcessor::filterChunks() {
     return true;
 }
 
-bool AudioProcessor::mergeChunks() {
-    ConfigManager &configManager = ConfigManager::getInstance();
-    std::string ffmpegPath = configManager.getFFmpegPath();
-
-    // Merge processed chunks with `crossfade`
-    std::string ffmpegMergeCommand = ffmpegPath + " -y ";
-    for (int i = 0; i < m_processedChunkPaths.size(); ++i) {
-        ffmpegMergeCommand += "-i \"" + m_processedChunkPaths[i] + "\" ";
-    }
-
+std::string AudioProcessor::buildFilterComplex() const {
     // Build filter complex, i.e. a set of instructions for FFmpeg (called filter graph)
     std::string filterComplex = "";
     int filterIndex = 0;
+
+    if (m_processedChunkPaths.size() < 2) {
+        return filterComplex;  // Return empty string if not enough chunks
+    }
+
     for (int i = 0; i < m_processedChunkPaths.size() - 1; ++i) {
         if (i == 0) {
             // Generate a `crossfade` for the first chunk pair (0 and 1)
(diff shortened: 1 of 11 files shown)


MediaProcessor/src/ConfigManager.cpp
```diff
@@ -29,4 +29,11 @@ std::string ConfigManager::getFFmpegPath() const {
     return m_config["ffmpeg_path"].get<std::string>();
 }
 
+int ConfigManager::getNumChunks() const {
```
nikhiljangra264, line 32: Added a New Funtion to get the num_chunks from config File

MediaProcessor/src/AudioProcessor.cpp
```diff
@@ -15,19 +15,40 @@ AudioProcessor::AudioProcessor(const std::string &inputVideoPath,
...
+    /*
+     * If the value is not well-defined or not computable, hardware_concurrency() may return 0.
+     * In that case, we'll fall back to DEFAULT_NUM_CHUNKS.
+     * We subtract 2 from the available hardware threads as a safety margin to avoid overloading the system.
+     */
+    ConfigManager &configManager = ConfigManager::getInstance();
+    int configNumChunks = configManager.getNumChunks();
+    int hardwareNumChunks = std::thread::hardware_concurrency();
+
+    if (configNumChunks >= 1 && configNumChunks <= hardwareNumChunks) {
```
omeryusufyagci, line 43: Please add an `isWithinRange` to `Utils`, so that this reads `if(isWithinRange)` which is much easier to reason about

MediaProcessor/src/AudioProcessor.cpp
```diff
@@ -15,19 +15,40 @@ AudioProcessor::AudioProcessor(const std::string &inputVideoPath,
...
+    std::cout << "INFO: using " << std::to_string(m_numChunks) << " Threads." << std::endl;
+
     /*
      * TODO: 
      * * numChunks (numThreads) should be configurable, as well as the overlap (which seems much less critical?)
      * * Need an ffmpeg utils to establish an API for extaction and probing
      */
 }
 
+int AudioProcessor::determineNumChunks() const {
```
omeryusufyagci, line 33: Please move this method to under `ConfigManager` and rename it to `getOptimalThreadCount`.
This new getter should call a private method `determineNumThreads`, where we determine this based off of hardware information. 

For this hardware information, use add a new `HardwareUtils` with a `getHardwareThreadCount()`. This will simply return the info via the `std::thread::hardware_concurrency()`.

Roughly:
```cpp
// in ConfigManager
unsigned  getOptimalThreadCount() {
    unsigned int configNumThreads = ... // get from config
    unsigned int hardwareNumThreads = MediaProcessor::HardwareUtils::getHardwareThreadCount();

    return determineNumThreads(configNumThreads, hardwareNumThreads);
}
```

This will help avoid coupling loosely related classes together.

MediaProcessor/src/AudioProcessor.cpp
```diff
@@ -15,19 +15,40 @@ AudioProcessor::AudioProcessor(const std::string &inputVideoPath,
                                const std::string &outputAudioPath)
     : m_inputVideoPath(inputVideoPath),
       m_outputAudioPath(outputAudioPath),
-      m_numChunks(DEFAULT_NUM_CHUNKS),
+      m_numChunks(determineNumChunks()),
```
omeryusufyagci, line 18: Please remove this from the member initialization. Instead, initialize it inside the constructor body by using the `ConfigManager` singleton via the new `getOptimalThreadCount`

config.json
```diff
@@ -2,5 +2,6 @@
     "deep_filter_path": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
     "downloads_dir": "downloads",
     "ffmpeg_path": "/usr/bin/ffmpeg",
-    "upload_folder": "uploads" 
+    "upload_folder": "uploads",
+    "num_chunks": 0
```
omeryusufyagci, line 6: Please rename it to `num_threads`. Users should not be concerned with how we manage our chunking wrt thread availability, and we may change how we chunk later on, and that shouldn't break the config or the user API. 

Also, it shouldn't default to 1 and not 0.

MediaProcessor/src/ConfigManager.cpp
omeryusufyagci: As explained below, please rename to `getNumThreadsValue`, as to avoid chunks in the name, which is internal to AudioProcessor, as well as to adhere to convention of value for key.

README.md
```diff
@@ -145,6 +145,16 @@ python3 app.py
 ```
 > **Note**: The server should indicate it is running on http://127.0.0.1:8080. Visit this address in your browser and submit a URL to begin.
 
+### Configuring the Number of Threads
+
```
omeryusufyagci, line 149: We can summarize this with a note at the end of prerequisites to say, that the user can adjust this via the `config.json` file.

omeryusufyagci (changes requested):
Thanks for the great work, and a solid start! I see some opportunity to improve certain parts of this. Please address these comments and we'll take another look at it.

config.json
nikhiljangra264: > Also, it shouldn't default to 1 and not 0.
I think we shouldn't place the num_threads in the config at starting. Beacuse most of the users are not concerned about the threads it uses.
so there is no point of default value.
users can add this setting if they want to limit the number of threads.
???

omeryusufyagci: You're right. We should then introduce a new field in the config `limit_thread_count` that defaults to `false` and rename `num_chunks` to `max_thread_count` which you can default to a low number, e.g. 4. 

`max_thread_count` is only relevant if `limit_thread_count` is `true`.

nikhiljangra264:
I have make the changes requested. Please check

omeryusufyagci:
> I have make the changes requested. Please check

@nikhiljangra264 I needed to push an update on main to unblock a few things. Could you please make sure your PR conforms? Details available in #27. Thanks a lot!

nikhiljangra264:
@omeryusufyagci  I have made Changes. Please look

omeryusufyagci (approved):
Nice, PR, thanks! Just did some minor refactors and ran it through the updated clang-format. 
Changed `json.config` fields to be more descriptive, they read better now. 
Also added a helper method in Utils.

Issue #18, opened 2024-10-01
Testing: Perform Initial Tests on Windows


omeryusufyagci:
We are looking to expand support to Windows users by providing an installable Windows release. Before we proceed to automate a Windows release pipeline, we need to confirm that the current project works well on a Windows environment.

This is a particularly good first issue to take, and a good opportunity to understand the inner workflow of the project. It will also pave the way to reach a greater audience. 

**Objectives**:
- Test Functionality on Windows:
  - Set up the project on a Windows machine by following the manual installation instructions.
  - Run through the core functionality (e.g., processing a video, extracting audio) and verify if everything works as expected.

- Identify Platform-Specific Issues:
  - Report any issues specific to Windows (e.g., file path problems, dependency issues, permission errors).
  - If possible, suggest solutions or fixes to make the project compatible.

- Provide Feedback:
  - Document what worked and what didn't in your testing environment (Windows version, any special setup you needed to do).
  - Suggest improvements to the documentation if there are any changes needed for Windows users.

IbrahimHamshari:
Can you assign me this issue ?

omeryusufyagci:
Hi @IbrahimHamshari, thanks for your interest! 

I've assigned the issue to you. Please take note of each problem you encounter and provide updates to our documentation to ensure a seamless manual installation experience for our Windows users. 

Feel free to reach out if you have any questions along the way

readysetlearn:
Hi @omeryusufyagci

I can't get this to build on Windows 10. I tried with Visual Studio and MSYS2. They both gave errors regarding the inclusion of nlohmann/json.hpp in ConfigManager.h. I tried installing the mingw-w64-x86_64-nlohmann-json package in MSYS2 but it didn't help.

omeryusufyagci:
Hi @readysetlearn thanks for reporting this. I'm expecting quite a few small issues like this to come up here. Let's revisit this with @IbrahimHamshari once his initial review is finished and proposes a set of manual installation instructions for Windows.

omeryusufyagci:
Trying to test on a Mac, I've also run into some problems and just updated the documentation to include details about the ffmpeg path and installation instructions for the JSON library. We can add the Windows instructions in the same collapsable section under Prerequisites. Now it seems to be fine on MacOS and Linux. 

Eventually we should provide an installation script, and detect the paths, update the config file ourselves, but for now let's try to document exhaustively.

IbrahimHamshari:
[...]

PR #22, opened 2024-10-04
feat: add windows support (MIDWAY)

IbrahimHamshari:
This is just a pull request to see why this error occurs when processing the video on Windows.

```
Error: Failed to process chunk with DeepFilterNet: C:\Users\acer\Desktop\Issues\fast-music-remover\uploads\chunks\chunk_4.wav
Command failed with return code 1:
'MediaProcessor' is not recognized as an internal or external command,
operable program or batch file.
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,6 +1,8 @@
 {
-    "deep_filter_path": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
+    "deep_filter_path_unix": "MediaProcessor/res/deep-filter-0.5.6-x86_64-pc-windows-msvc.exe",
+    "deep_filter_path_windows": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
     "downloads_dir": "downloads",
-    "ffmpeg_path": "/usr/bin/ffmpeg",
-    "upload_folder": "uploads" 
+    "ffmpeg_path_unix": "/usr/bin/ffmpeg",
+    "ffmpeg_path_windows": "ffmpeg",
+    "upload_folder": "uploads"
 }
--- a/MediaProcessor/src/AudioProcessor.cpp
+++ b/MediaProcessor/src/AudioProcessor.cpp
@@ -18,8 +18,8 @@ AudioProcessor::AudioProcessor(const std::string &inputVideoPath,
       m_numChunks(DEFAULT_NUM_CHUNKS),
       m_overlapDuration(DEFAULT_OVERLAP_DURATION) {
     m_outputDir = std::filesystem::path(outputAudioPath).parent_path().string();
-    m_chunksDir = m_outputDir + "/chunks";
-    m_processedChunksDir = m_outputDir + "/processed_chunks";
+    m_chunksDir = (std::filesystem::path(m_outputDir) / "chunks").string();
+    m_processedChunksDir = (std::filesystem::path(m_outputDir) / "processed_chunks").string();
 
     /*
      * TODO: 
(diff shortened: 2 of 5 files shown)


config.json
```diff
@@ -1,6 +1,8 @@
 {
-    "deep_filter_path": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
+    "deep_filter_path_unix": "MediaProcessor/res/deep-filter-0.5.6-x86_64-pc-windows-msvc.exe",
```
omeryusufyagci, line 2: Here you have the unix and windows paths inversed and the Windows path has the wrong slashes. Corrected file:

```json
{
    "deep_filter_path_unix": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
    "deep_filter_path_windows": "MediaProcessor\\res\\deep-filter-0.5.6-x86_64-pc-windows-msvc.exe",
    "downloads_dir": "downloads",
    "ffmpeg_path_unix": "/usr/bin/ffmpeg",
    "ffmpeg_path_windows": "C:\\ffmpeg\\bin\\ffmpeg",
    "upload_folder": "uploads"
}
```

ffmpeg path will be different for you

omeryusufyagci:
Thanks for the start on this! I took a look and was able to reproduce the problem you shared in #18, but it turned out to be a config issue. The MediaProcessor didn't compile for me as is, I had to adjust the CMakeLists to include a fetch for the json, and handle pthreads for unix only. After these the filtering worked fine. 

Then I ran into an issue with the `mergeChunks()`. It's a Windows-specific problem related to the quotes on the command itself causing it to not get parsed correctly. The same seems to work from the terminal.

I also ended up refactoring the entire core to eliminate strings for file paths, and strictly use `std::filesystem::path` everywhere. 

What I propose to do is to limit the scope of this PR to the work you've done for the `ConfigManager`, including the corrected `config.json` and cleaned version of the `app.py`. Let's merge that and I'll push my refactor along with the CMakeChanges and the `CommandBuilder` I wanted to add. Then, you can then tackle the remaining issues if you're up for it.

For this to work though, it can't break the Unix/Linux version. So please test there, and only add non-breaking changes. How does this sound?

omeryusufyagci (changes requested):

IbrahimHamshari:
So for some reason, it works now on both Linux (using docker) and Windows right now.
Here are the steps that worked for me, 
First I installed  GCC using MSYS2 with this command:

`pacman -S mingw-w64-x86_64-toolchain base-devel`

Then install Cmake using this command :

`pacman -S mingw-w64-x86_64-cmake`

Then Installed the nlohmann/json.hpp by using this command:

`pacman -S mingw-w64-x86_64-nlohmann-json`

if the ffmpeg is properly installed in the `C:\ffmpeg\bin` it should work.

omeryusufyagci:
> So for some reason, it works now on both Linux (using docker) and Windows right now. Here are the steps that worked for me, First I installed GCC using MSYS2 with this command:
> 
> `pacman -S mingw-w64-x86_64-toolchain base-devel`
> 
> Then installed Cmake using this command :
> 
> `pacman -S mingw-w64-x86_64-cmake`
> 
> Then Installed the nlohmann/json.hpp by using this command:
> 
> `pacman -S mingw-w64-x86_64-nlohmann-json`
> 
> if the ffmpeg is properly installed in the `C:\ffmpeg\bin` it should work.

The video playback works on the UI? I had an issue with the ffmpeg command for mergeChunks yesterday. I'll take a look

omeryusufyagci:
Hi @IbrahimHamshari, I refactored the core to strictly use `std::filesystem::path` everywhere, and introduced a command building interface. It was painful to work with the hardcoded, in-place commands. 

Could you please update this to conform to these changes as well as to target the PR for the new branch `windows-support`?

Thanks a lot!

IbrahimHamshari:
I'll close this pull request and then open another one which has the complete thing.

PR #29, opened 2024-10-07
Windows support

IbrahimHamshari:
### Steps:

1. Install the gcc using the MSYS2
`pacman -S mingw-w64-x86_64-toolchain base-devel`

2. Install Cmake using MSYS2
`pacman -S mingw-w64-x86_64-cmake`

3. Install the nlohmann/json.hpp (I don't know if this is needed anymore.)
`pacman -S mingw-w64-x86_64-nlohmann-json`

4. Install ffmpeg and put it in the environment PATH variable.

5. Run Cmake using the MSYS2 in the build directory
`cmake -G "MSYS Makefiles" ..`

6. Run the make command in the build directory
`make`

7. Run the application using `python app.py`

Tested it in the windows environment, and the Linux environment (using docker).


omeryusufyagci:
Hi @IbrahimHamshari,

Many thanks for the changes and instructions, it works! However, I'm not too keen on being fully dependent on the msys env. Requiring manual exports and having to do the entire workflow within this shell feels limiting. For a Windows release, I'd prefer a solution that doesn't rely on msys and works directly on a standard Windows setup.

I don't have much experience with native Windows development myself. Do you have any ideas for an alternative? If we can't find a better solution, we should at least provide clear documentation to ensure it works on a clean system.

This branch is a bit outdated, so I’ll update it and plan to merge over the weekend unless you have any other updates.

Thanks again, great to have something working on Windows!

IbrahimHamshari:
@omeryusufyagci I'm trying to make it work without depending on the msys env. I do think I'm making some progress.

IbrahimHamshari:
Unfortunately, I wasn't able to do that without the msys env as it's more complex to not use msys than to use it, so I think it would be better to just use it.
@omeryusufyagci Should I just update the CONTRIBUTING.md or both the CONTRIBUTING.md and the README.md ?

omeryusufyagci:
Well, thanks for trying @IbrahimHamshari! I also don't have a better solution at the moment, but for the current stage of the project, I think this is acceptable. I'd rather shift our focus towards providing a stable Windows binary.  

Since these are development details, updating the contributing should be enough. Thank you!

Issue #8, opened 2024-09-29
Testing: Add Unit Tests for `MediaProcessor` using Google Test


omeryusufyagci:
To eliminate the need for manual testing and verification of the `MediaProcessor` after every change, unit tests should be added to ensure consistent behavior. Given the nature of the project, this requires appropriate input/output media files to thoroughly test the core functionality of the `MediaProcessor`.

**Objectives**:
1. Find/Generate suitable test I/O for the `MediaProcessor`: 
Identify or generate media files that can be used for testing. Ideally, these should be short (a few seconds long) clips covering different use cases (e.g., video with background music, video with pure vocals, noisy background).

2. Implement Tests Using Google Test:
Add tests for critical operations like audio extraction, filtering, and merging back the processed audio with video.

nikhiljangra264:
I would Love to work on this. Can you assign me this?

omeryusufyagci:
Hi @nikhiljangra264, would be great, thanks!

PR #36, opened 2024-10-08
Testing: Add Unit Tests for MediaProcessor using Google Test #8

nikhiljangra264:
#8 

## Summary
This PR introduces the Google Test (gtest) framework, a `ConfigManagerTester.cpp` test suite, and restructures the project with separate `CMakeLists.txt` files for source, tests, and the main project.

## Changes

1. **Modular CMakeLists Structure**
   - Created three separate `CMakeLists.txt` files for better project organization:
     - `src/CMakeLists.txt`: Handles the build configuration for the source code.
     - `tests/CMakeLists.txt`: Configures the build for test-related files, including the integration of `gtest`.
     - `CMakeLists.txt`: The main CMake file, fetch content and control the two other files.

2. **Add gtest Framework**
   - Integrated `gtest` into the `tests/CMakeLists.txt` for unit testing.
   
3. **Add ConfigManagerTester**
   - Introduced a `ConfigManagerTester.cpp` in the `tests` directory.
   - Currently check for proper reading of config.json file by ConfigManager.cpp

4. **Add ConfigFileCreator**
   - Make a class TempConfigFile responsible for creating temporary config file to be used by tests


omeryusufyagci:
Hey @nikhiljangra264 can you target this to `8-testing-add-unit-tests-for-mediaprocessor-using-google-test`, thanks!

nikhiljangra264:
Done!

MediaProcessor/tests/ConfigFileCreator.h
```diff
@@ -0,0 +1,31 @@
+#ifndef CONFIGFILECREATOR_H
+#define CONFIGFILECREATOR_H
+
+#include <fstream>
+#include <nlohmann/json.hpp>
+#include <stdexcept>
+
+namespace MediaProcessor {
+/// @brief Create a temporary configuration file from a JSON object.
```
omeryusufyagci, line 9: Nice to see this! I was planning to start using the doxy comment strings in the core. So, good to start with the new code. Please use the `/**/` version though.

omeryusufyagci (changes requested):
@nikhiljangra264 thanks for the great start! I left some comments, and I'm open to discuss if you have any feedback on them. 

Could you also please setup tests for other classes as well, as the underlying issue addresses them all? 
Even if the coverage is low at first, it would be great to have the complete structure in place before merging this.

Thanks again! This is really helpful, and I appreciate your effort!

MediaProcessor/CMakeLists.txt
```diff
@@ -10,21 +10,23 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-include_directories(include)
-include_directories(${CMAKE_SOURCE_DIR}/third_party/nlohmann)
+# Fetch Google Test
+include(FetchContent)
 
+# Set the policy to NEW
+cmake_policy(SET CMP0135 NEW)
```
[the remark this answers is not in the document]
nikhiljangra264: This set fetch content policy to new. This policy is introduced in CMake 3.19. I use it in case It didn't fetch GTest. But i removed it in future commits. ( yet to be pushed )

nikhiljangra264:
I added most of the test suits. Please tell me if any remains

MediaProcessor/tests/AudioProcessorTester.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // Load the config
+    ConfigManager &configManager = ConfigManager::getInstance();
+    ASSERT_TRUE(configManager.loadConfig(testConfigFile.getFilePath()))
+        << "Unable to Load TestConfigFile";
+
+    std::filesystem::path testAudioPath = testOutputDir / "test_output_audio.wav";
+    AudioProcessor audioProcessor(testVideoPath, testAudioPath);
+
+    // Test the isolateVocals function
+    EXPECT_EQ(audioProcessor.isolateVocals(), true);
```
omeryusufyagci, line 62: This is nice, you're checking if it finishes correctly, but this doesn't tell us if the filter worked correctly. One of the annoying parts currently is that after every change I have to manually inspect that the the filter still works properly. 

It would be such a time saver if you could add a test case that compares the isolated vocals with a known audio track. We can add another audio track to TestMedia which will be the filtered version of `test_audio.waw`, and we can compare the 2 signals with some tolerance to confidently say the filter still works. 

Could you try to see if this is feasible?

MediaProcessor/tests/ConfigFileCreator.h
```diff
@@ -0,0 +1,44 @@
+#ifndef CONFIGFILECREATOR_H
+#define CONFIGFILECREATOR_H
+
+#include <filesystem>
+#include <nlohmann/json.hpp>
+#include <stdexcept>
+
+namespace MediaProcessor::UnitTesting {
```
omeryusufyagci, line 8: bit of a nitpick, let's rename the NS to `MediaProcessor::Tests`

MediaProcessor/tests/ConfigManagerTester.cpp
```diff
@@ -0,0 +1,39 @@
...
+    TestConfigFile testConfigFile("testConfig.json", jsonContent);
+
+    // Load the config
+    ConfigManager &configManager = ConfigManager::getInstance();
+    bool loadSuccess = configManager.loadConfig(testConfigFile.getFilePath());
+
+    // tests
+    EXPECT_TRUE(loadSuccess);
+    EXPECT_EQ(configManager.getDeepFilterPath(),
+              "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl");
```
omeryusufyagci, line 34: to avoid hardcoding it here, can you access it through the json map you have above?

omeryusufyagci (changes requested):
This looks like it'll be a nice start for us! Apart from 2 minor points, the only big change I'd love to have is a proper functionality test for the filter. This is more so an integration test rather than an isolated unit test, but it's possibly the most important test we must do before each merge/push. 

Since this will form the initial test structure, I am fine with adding this one integration test here, and later we can look for better separation for CI. 

Thanks for the great work!

MediaProcessor/tests/ConfigFileCreator.h
nikhiljangra264: ok

MediaProcessor/tests/AudioProcessorTester.cpp
nikhiljangra264: I am looking to add it but first i have to make sure it generates similar audio if generated with different number of threads or different settings. so i skip this for now. i will add it

MediaProcessor/tests/testUtils.cpp
```diff
@@ -0,0 +1,103 @@
...
+    sf_close(sndFile1);
+    sf_close(sndFile2);
+
+    return compareBuffersWithTolerance(buffer1.begin(), buffer1.end(), buffer2.begin(),
+                                       buffer2.end(), tolerance);
+}
+
+template <typename Iter>
+bool compareBuffersWithTolerance(Iter begin1, Iter end1, Iter begin2, Iter end2, double tolerance) {
+    while (begin1 != end1 && begin2 != end2) {
```
omeryusufyagci, line 91: We should make this readable. You could use `std::equal` with a comparator lambda or extrract it as a util, but the code here should read plainly

omeryusufyagci (changes requested):
Hi @nikhiljangra264, thanks a lot! You're basically building the testing structure for our core and it looks like we're almost there for the first version. Please address my comments and let's take another look!

MediaProcessor/src/ConfigManager.h
```diff
@@ -53,6 +53,8 @@ class ConfigManager {
      */
     unsigned int determineNumThreads(unsigned int configNumThreads,
                                      unsigned int hardwareNumThreads);
+    template <typename valueType>
```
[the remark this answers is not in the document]
nikhiljangra264: You should've mentioned on one only :joy:

MediaProcessor/src/ConfigManager.cpp
```diff
@@ -19,29 +19,45 @@ bool ConfigManager::loadConfig(const fs::path& configFilePath) {
         std::cerr << "Error: Could not open " << configFilePath << std::endl;
```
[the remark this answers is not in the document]
nikhiljangra264: ok

Issue #32, opened 2024-10-07
Feature: Add `DeepFilterCommandBuilder` to `MediaProcessor`


omeryusufyagci:
Implement a `DeepFilterCommandBuilder` class to establish a high-level API for constructing `DeepFilterNet` commands, extending the CommandBuilder base class. Methods such as `setInputAudio()` and `setNoiseReductionLevel()` should abstract the command details.

This will make the filtering process more maintainable and user-friendly, eliminating the need for users to understand command-level specifics. Will be needed by the frontend for v1.0.s

```cpp
DeepFilterCommandBuilder cmd;
cmd.setInputAudio(inputAudioPath);
cmd.setOutputAudio(outputAudioPath);
cmd.setNoiseReductionLevel(0.5);
// ...
```
See discussion in #4 for more context.

A good first issue as it's a direct extension of an existing class, requiring minimal understanding of the full project.

georgeolson92:
Hey there! I wanted to see if you need some help getting this implemented. Would be happy to help!

omeryusufyagci:
Hi @georgeolson92, it would be great, thanks a lot! Feel free to reach out if you have questions/feedback along the way.

omeryusufyagci:
This was closed by #42

PR #42, opened 2024-10-14
Feature: Add DeepFilterCommandBuilder to MediaProcessor

georgeolson92:
This is my first time working in the code base (and with anything related to deep filtering), hope this will address the new feature implementation in issue #32! Let me know what needs to be adjusted, or if there is anything else I should add!


omeryusufyagci (changes requested):
Hi @georgeolson92, thanks for the changes! I added a few comments addressing minor points. 

Could you also take a look at other flags that may be interesting to have, for a better coverage of what's available?

For instance, I remember using a post filter with a post filter beta. There was also a field for adjusting the filtering limits in dB. 
We'll be exposing the filtering behavior to the end-user, so these can come in handy. 

I haven't tested the `--noise-reduction` flag so far though, maybe it works great on its own? 

Thanks a lot for taking another look!

MediaProcessor/src/DeepFilterCommandBuilder.cpp
```diff
@@ -0,0 +1,41 @@
...
+    return *this;
+}
+
+DeepFilterCommandBuilder& DeepFilterCommandBuilder::setNoiseReductionLevel(double level) {
+    if (level < 0.0 || level > 1.0) {
+        throw std::invalid_argument("Noise reduction level must be between 0.0 and 1.0");
+    }
+    this->noiseReductionLevel = level;
+
+    addArgument("--noise-reduction");
```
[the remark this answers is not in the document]
georgeolson92: Thanks for letting me know, I have updated this in the code!

MediaProcessor/src/DeepFilterCommandBuilder.cpp
```diff
@@ -0,0 +1,41 @@
...
+    addArgument(std::to_string(level));
+    return *this;
+}
+
+std::string DeepFilterCommandBuilder::build() const {
+    if (inputAudioPath.empty() || outputAudioPath.empty()) {
+        throw std::runtime_error("Input and output audio paths must be specified.");
```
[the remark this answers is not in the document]
georgeolson92: Implemented this in new commit!

MediaProcessor/src/DeepFilterCommandBuilder.cpp
```diff
@@ -0,0 +1,41 @@
+#include "DeepFilterCommandBuilder.h"
+
+#include <stdexcept>
+#include <sstream>
+
+namespace MediaProcessor {
+
+DeepFilterCommandBuilder& DeepFilterCommandBuilder::setInputAudio(const std::string& inputAudioPath) {
+    this->inputAudioPath = inputAudioPath;
```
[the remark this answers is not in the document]
georgeolson92: Went ahead and made these changes and also updated the header file as well! Thanks for the heads up!

georgeolson92:
Made updates based on recommendations from review, let me know how everything looks! Thank you for all of the helpful feedback!

MediaProcessor/src/DeepFilterCommandBuilder.cpp
```diff
@@ -0,0 +1,43 @@
+#include "DeepFilterCommandBuilder.h"
+#include <stdexcept>
+#include <sstream>
+
+namespace MediaProcessor {
+
+DeepFilterCommandBuilder& DeepFilterCommandBuilder::setInputAudio(const std::string& inputAudioPath) {
+    m_inputAudioPath = inputAudioPath;
+    addFlag("--input");
```
omeryusufyagci, line 9: Currently we are passing the input file path as the argument [here](https://github.com/omeryusufyagci/fast-music-remover/blob/e04313276c5149d66853c8864a546eeda68e4fa2/MediaProcessor/src/AudioProcessor.cpp#L153) without an `--input` flag, is this supported as well? 
If not we can simply pass  this path via the `addArgument()` method. 
Could you please check?

MediaProcessor/src/DeepFilterCommandBuilder.cpp
```diff
@@ -0,0 +1,43 @@
...
+    return *this;
+}
+
+DeepFilterCommandBuilder& DeepFilterCommandBuilder::setNoiseReductionLevel(double level) {
+    if (level < 0.0 || level > 1.0) {
+        throw std::invalid_argument("Noise reduction level must be between 0.0 and 1.0");
+    }
+    m_noiseReductionLevel = level;
+
+    addFlag("--noise-reduction");
```
omeryusufyagci, line 27: Here and in the line below, please use the correct overload of `addFlag` that allows passing a flag with a value. This will improve consistency. Could you update this and other similar occurrences?

omeryusufyagci (changes requested):
Hi @georgeolson92, thanks for the changes! One of the mandatory checks failed as the code was not `clang-format`'ed. I also noticed a couple of other things for you to check. 

Could you please address the 2 points and run the `clang-format`? Should be good to go after. Many thanks! 

Edit: we provide a `.clang-format` at the project root, in case you havent' seen it yet

MediaProcessor/src/DeepFilterCommandBuilder.cpp
georgeolson92: Hey, thanks for pointing this out, it looks like it is not necessary so we can go ahead and just use the addArgument() method!

MediaProcessor/src/DeepFilterCommandBuilder.cpp
georgeolson92: Will do, going to push this update right away!

georgeolson92:
Thanks for the review, made requested updates and ran clang-format, let me know how it's looking!

omeryusufyagci (commented):
Hi @georgeolson92, many thanks for the changes! 

The only important thing that was missing was a method to set the `--compensate-delay` flag for the filter. This ensures audio remains in sync with the original media after filtering. 

I also made some minor refactors for consistency across the codebase.

One issue we'll need to address in the future is that this filter does not currently support specifying an input flag, and it expects the input file to be provided after the options. This approach is not ideal for a builder, as it forces users to follow a specific sequence of operations. I didn't find a straightforward solution to fix this without compromising const-correctness, so we'll revisit it later.

We will most likely structure the DeepFilter related classes similar to what we discussed at #45. Please let me know if you'd be interested in continuing this work further. 

Thanks again!

georgeolson92:
Hi @omeryusufyagci, I can certainly continue working on the Deep Filter classes, if you have any issues open feel free to assign me and I can create a new branch for that!

omeryusufyagci:
Hi @georgeolson92, fantastic to hear! : ) We have #65, and #52 that are DeepFilter specific. I am not able to assign you directly -for some reason can't find you on the list. If you could leave a comment on these I can assign them to you. 

As usual, I am happy to clarify anything that may come as we progress. 

Thanks a lot!

PR #51, opened 2024-10-21
Add `FFmpegCommandBuilder` class

### LINKED ISSUE #31: Feature: Implement `FFmpegCommandBuilder` to `MediaProcessor`
Implement an `FFmpegCommandBuilder` class to extend the existing `CommandBuilder`, to establish a high-level API, with methods like overwriteOutput(), setInput(), and for other settings. 

This will simplify `FFmpegController` usage by improving readability and reducing direct flag usage in the codebase. It will also make the implementation more user-friendly, removing the need to know FFmpeg command details.

```cpp
FFmpegCommandBuilder cmd;
cmd.overwriteOutput();
cmd.setInput(inputVideoPath);
cmd.setSampleRate(m_audioConfig.sampleRate);
// ...
```
For more context, see the discussion in #4.

A good first issue as it's a direct extension of an existing class and requires minimal understanding of the full project.

andcscott:
Hi @omeryusufyagci 

This PR closes #31 and makes the following changes:

1. Rebase the current branch to include updates from the main branch
2. Add getters for the codec value maps in `FFmpegSettingsManager`
3. Move the `enumToString` method from `FFmpegSettingsManager` to `Utils`
4. Update `CommandBuilder` for method chaining with `FFmpegCommandBuilder`
5. Add the `FFmpegCommandBuilder` class


omeryusufyagci (changes requested):
Hi @andcscott, thank you for these changes and the commit hygiene -makes it really easy to review these! 

So,  reviewing this I realized that making the settings manager a singleton was not ideal at all! It doesn't provide much value, and if anything, it can easily become detrimental and tedious to work with it. 

So I quickly revised it to not be that! I didn't move the enumToString util as you handled that, but this would require some changes on this PR (to the better I promise!)

You should take a reference of the settings manager now in the ctor. 
I would also advise to take the ffmpeg path from ConfigManager and updating a member var in the ctor as well. 

I would also like you to handle the setInputFile in the builder. A settings manager should not deal with I/O file management. It makes sense to make that a build stage. 

Finally, I'd like you to think about adding some checks where sensible, and throwing if we don't find what we expect.  These can be missing files, settings etc. We have to document clearly where we throw in the header. 

I just merged my PR to update the settings manager. Please take a look and let me know if you have any comments or if anything is unclear.

Many thanks!

andcscott:
Hey there @omeryusufyagci thanks for reviewing another PR!

No worries about the singleton problem, it totally makes sense. About to push some changes, let me know what you think.

Summary:

1. Ctor now takes a reference to the settings manager and stores in a private var
2. Methods for input/output paths now take the path and also store them in vars
3. Added error checking

The error checking is probably my biggest concern with this push, I wasn't sure whether to do it method by method or all at once in the build method. For now I did a little of both where it seemed to make sense, trying to be consistent with the new deep filter builder output. I also took the liberty of ensuring a sane audio sample rate, but let me know what you think.

MediaProcessor/src/FFmpegCommandBuilder.cpp
```diff
@@ -0,0 +1,96 @@
...
+    addFlag("-c:a", codec);
+
+    return *this;
+}
+
+FFmpegCommandBuilder& FFmpegCommandBuilder::addAudioSampleRate() {
+    int sampleRate = m_ffmpegSettings.getAudioSampleRate();
+
+    // Nyquist-Shannon theorem
+    if (sampleRate <= 40000) {
```
omeryusufyagci, line 50: I'll explain in the review comment, but let's remove checks done on setting fields. Anything that doesn't come from the settings manager, we should still check.

MediaProcessor/src/FFmpegCommandBuilder.h
```diff
@@ -0,0 +1,111 @@
...
+     * @brief Adds overwrite flag
+     *
+     * @return A reference to the current object for method chaining
+     */
+    FFmpegCommandBuilder& addOverwrite();
+
+    /**
+     * @brief Adds input flag and input file path
+     *
+     * @param inputFile The input file path
```
omeryusufyagci, line 36: For simple variables like these, we don't need to add a `@param` entry. For large headers it can introduce some clutter.

MediaProcessor/src/FFmpegSettingsManager.h
```diff
@@ -0,0 +1,81 @@
...
+ * Provides an interface for setting and retrieving configuration options,
+ * including file paths, codecs, and other processing parameters.
+ */
+class FFmpegSettingsManager {
+   public:
+    FFmpegSettingsManager();
+
+    // Global Setters
+    void setOverwrite(bool overwrite);
+    void setInputFile(const std::string& inputFile);
```
omeryusufyagci, line 30: This, setOutputFIle and their getters can now be removed from here

MediaProcessor/src/FFmpegSettingsManager.h
```diff
@@ -0,0 +1,81 @@
+#ifndef FFMPEGSETTINGSMANAGER_H
+#define FFMPEGSETTINGSMANAGER_H
+
+#include <string>
+#include <unordered_map>
+
+/**
```
omeryusufyagci, line 7: You can also remove this todo as you've fixed this

omeryusufyagci (changes requested):
This builder itself is pretty much ready. Checking for essential elements, e.g. file paths, directly at `build()` is a good approach. 

The settings should have some sensible defaults though. However, these should come from the settings manager. Similarly, checks for invalid settings should happen there as well. So, the builder can and should assume that the settings it receives are valid. I'll try to fix this as soon as I can, but it's not a reason to block this PR.

In the meantime, could you update the builder to remove the checks on settings?
I also requested a couple of minor changes on the FFmpegSettingsManager. Since we added the I/O directly to the builder, we should remove them from the settings manager within this PR for clarity. 

Otherwise looks great, thanks a lot!

andcscott:
Sounds good, just pushed the requested changes. If everything looks OK I'll move forward with FFmpegControllerBuilder. Thanks!

omeryusufyagci (approved):
Looks good, thanks!

Issue #34, opened 2024-10-07
Improvement: Backend Refactor for Typing, Tests, Modularity and Pathlib


omeryusufyagci:
Current `app.py` will be reorganized under `backend/`, with the following objectives:

- **Typing**: Add type hints throughout the backend
- **Unit Tests**: Develop unit tests for all utility and processing functions.
- **Response Handling**: Add a `ResponseHandler` class in `backend/response_handler.py` to handle all communication bwn BE-FE, BE-core using consistent JSON responses.
- **Replace os with pathlib**
- **Reorganize the code under `backend/`**:
  - Move `Utils` to `backend/utils.py` 
  - Move `MediaHandler` to `backend/media_handler.py`  
  - Keep `app.py` for route definitions and Flask app initialization.
  
This is a big step going towards v1.0 and offers a relatively straightforward first issue for contributors looking to make an impactful contribution to the project.

wenayy:
assign me

wenayy:
you can check it have made an pr #39

prakash-2002-ramanathan:
Is this issue open? Shall I open a PR?

omeryusufyagci:
Hi @prakash-2002-ramanathan, thanks for your interest. Indeed, it's open and much needed! 

Would be great if you could open a PR!

prakash-2002-ramanathan:
Hey, I have done all of the work, but not quite sure what you mean by "make the Backend to Core communication Consistent JSON "?
Like You want to start a server in C++ to handle JSON responses or pass the data in JSON format to the Media Handler subprocess ?

omeryusufyagci:
Thanks for the question @prakash-2002-ramanathan, 

We should ensure consistency across the application, using json responses between all layers. Currently the MediaProcessor writes to stdout and the backend is parsing directly these traces. Instead, the core should return a json object. 

Likewise, anything between the frontend and backend should be managed through json objects for consistency. 

Does this answer your question?

omeryusufyagci:
Well, we have an open issue to add a proper logger, e.g. spdlog. Until that comes, we're still writing plain text to cout/cerr. 

In the future we'll probably use gRPC for inter-process communication (IPC), but at this stage of the project I'd rather get something significantly less error prone working first. In that sense, for now, we can dedicate cerr for logging, leaving cout isolated for our IPC. 

Let's make sure the json structure stays consistent, so that we can write generic parsers. 
I'll suggest something like: 
```json
{
[...]

PR #54, opened 2024-10-23
Reorganize be v1.0

prakash-2002-ramanathan:
Hello @omeryusufyagci 

This PR addresses #34. 
Please see if the current work is satisfactory

Possible issues:
- I have changed the `config.json`
- Changed some names
- Changed Logs
- JSON message structure
- Added new variable `config_file_path` in MediaHandler `main` to pass the `config.json` file directory. 

I have not removed the original app.py. Not updated the README. Have not updated the requiremets.pip, some minor work is pending.


omeryusufyagci (changes requested):
Hi @prakash-2002-ramanathan, thanks for this PR!

I skimmed through the changes, and I'd like to touch on a few points:
1. CLI nature of core: For the moment, we need to keep this capability. There is an open issue assigned to build a proper CLI for the core, but until that’s addressed, it needs to remain independently usable.
2. Keeping `main` clean: `main` should serve strictly as a simple entry point to the application, without containing any processing logic. Please encapsulate these into a class like `CommunicationHandler` and simply use its methods on main. Here's an example of what I had on my mind:

This:
```cpp
    if (argc != 2) {
        std::cerr << "Usage: " << argv[0] << " <video_file_path>" << std::endl;
        return 1;
    }
```
would become:
```cpp
    if (argc != 2) {
        CommunicationHandler::handleError("Usage: ...");
    }
```
and then within `handleError()` you'd call a member function to generate a json response and exit with a sensible code. This would allow us to abstract communication handling and improve over time while keeping the public API stable.  

I understand this is not true IPC, but as I explained in #34, the immediate goal is to eliminate the existing error-prone implementation. Improving how we handle responses within core will already be a big step in the right direction. 

3. Config file path changes: I didn't quite understand why adding the `../` before MediaProcessor paths were needed there. Could you clarify why this was necessary?

prakash-2002-ramanathan:
@omeryusufyagci, The file path in python is relative to the python file. But since the MediaHandler is outside the backend directory. We are going to the previous directory where the Mediahandler is present. This seemed like a simple fix

omeryusufyagci:
Hi @prakash-2002-ramanathan, I wanted to check with you how you're progressing.

Although I appreciate the effort directly on the core, #34 mainly addresses the backend, where additions like the response handler will be ready in place, pending an update of the core for full integration. 

You can still tackle the update of the core, but I'd propose to do it on a separate PR, and focus here fully on the backend improvements. The backend should function the same as it does right now, while the necessary handlers implemented and ready for the integration. 

Tests are a big part of this PR, and would be quite impactful as we get ready to make larger changes. 
Please

prakash-2002-ramanathan:
Hi @omeryusufyagci 
Yeah, that sounds good. I have already implemented changes in the core. 
Is that fine, just over 50 lines. 
I have written communication handlers for the core, but not used them
I have written communication handlers for the BE, and have used them. Except for the communication with the core

omeryusufyagci:
Hi @prakash-2002-ramanathan, that's great to hear! 

I would still like you to submit your work on the core as a separate PR though. Let's evaluate the work you've done on the backend, finalize, merge and then take a look at your work on the core separately. This is also to expedite the addition of backend tests in preparation of our first release. 

Finally, could you please target this PR to the dedicated branch opened under #34? Thanks!

Looking forward to see the changes!

prakash-2002-ramanathan:
Hi @omeryusufyagci ,  have stashed all the changes made in the core, only the changes in backend have been pushed.
 The CLI nature of the core and backend communication is currently preserved. 
 Fixed the imports and removed the old `app.py`

You can run the new app by going to the backend directory and then running app.py.

backend/media_handler.py
```diff
@@ -0,0 +1,82 @@
...
+            result = subprocess.run(
+                [str(base_directory / "MediaProcessor/build/MediaProcessor"), str(video_path), config_path], capture_output=True, text=True
+            )
+
+            if result.returncode != 0:
+                logging.error(f"Error processing video: {result.stderr}")
+                return None
+            
+            #To print all the logs which have been moved to stderr
+            print(result.stderr)
```
omeryusufyagci, line 65: please use the logger with appropriate levels instead of raw printing

backend/media_handler.py
```diff
@@ -0,0 +1,82 @@
...
+                    if processed_video_path.startswith('"') and processed_video_path.endswith('"'):
+                        processed_video_path = processed_video_path[1:-1]
+                    processed_video_path = str(Path(processed_video_path).resolve())
+                    logging.info(f"Processed video path returned: {processed_video_path}")
+                    return processed_video_path
+
+            return None
+        except Exception as e:
+            logging.error(f"Error running C++ binary: {e}")
+            return None
```
omeryusufyagci, line 82: As part of the backend refactor, please address the 2 todo messages in the code. Mainly we want to organize common utils under a utils class/file.

backend/response_handler.py
```diff
@@ -0,0 +1,36 @@
...
+        """Return a consistent JSON response for success."""
+        response = {
+            "status": "success",
+            "message": message,
+            "data": data or {}
+        }
+        return jsonify(response), 200
+
+    @staticmethod
+    def error(message: str, status_code: int = 400) ->tuple[Response, int]:
```
omeryusufyagci, line 18: I believe here the return type should be simply a Response type object. Flask understands it carries an error code, but it's not necessary to hint `int` on the return type.

backend/response_handler.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @staticmethod
+    def success(message: str, data: dict = None) ->Response:
+        """Return a consistent JSON response for success."""
+        response = {
+            "status": "success",
+            "message": message,
+            "data": data or {}
+        }
+        return jsonify(response), 200
```
omeryusufyagci, line 15: method names should be descriptive. Please use `generate_success_response()`, and similar for others. Thinking from the users perspective can help; `ResponseHandler.succes()` vs `ResponseHandler.generate_success_response()`

backend/response_handler.py
```diff
@@ -0,0 +1,36 @@
...
+    @staticmethod
+    def core_data_passer(message: str, data: dict = None) -> str:
+        """Return a JSON-formatted response for core components with provided data."""
+        core_response = {
+            "status": "success",
+            "message": message,
+            "data": data or {}
+        }
+        return dumps(core_response)
+    #You can only use jasonify for http communication to send the data as a string use json.dumps
```
omeryusufyagci, line 36: Could you make this a bit more clearer? Even for me it's abit difficult to understand what our intention is with this method. Is this a bidirectional serializer, or?.. Both in the method name and the docstring we should be able to immediately understand the purpose of this method.

config.json
```diff
@@ -1,8 +1,8 @@
 {
-    "deep_filter_path": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
-    "deep_filter_tarball_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx.tar.gz",
-    "deep_filter_encoder_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/enc.onnx",
-    "deep_filter_decoder_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/df_dec.onnx",
+    "deep_filter_path": "../MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
+    "deep_filter_tarball_path": "../MediaProcessor/res/DeepFilterNet3_ll_onnx.tar.gz",
+    "deep_filter_enoder_path": "../MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/enc.onnx",
+    "deep_filter_decoder_path": "../MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/df_dec.onnx",
```
omeryusufyagci, line 5: This should be handled on the backend side. The config represents the path relation of these files wrt the project root.

omeryusufyagci (changes requested):
Hey @prakash-2002-ramanathan, thanks for the changes! 

We're on the right track, but there are still a few things to address. Also your changes on the core are still visible on the PR. 

I only reviewed the backend changes, and will review the tests once the backend part is finished. 

In addition to the review comments, I'd also like you to keep the `app.py` at project root, but this will be a minimal entry point for the backend. So 2-3 lines just to instantiate thje backend service and run it. This is to ensure we keep our current ease of use. Please feel free to make the necessary changes to faciliatate this as well. 

Many thanks!

prakash-2002-ramanathan:
Hi @omeryusufyagci , fixed all your requested changes. Please have a look

backend/response_handler.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @staticmethod
+    def generate_error_response(message: str, status_code: int = 400) ->Response:
+        """Return a consistent JSON response for errors."""
+        response = {
+            "status": "error",
+            "message": message,
+            "data": {}
+        }
+        return jsonify(response), status_code
```
omeryusufyagci, line 25: Here please add a `generate_response()` function to hold shared behavior, that is, generating a response. 
That new method should take in any data available with the status code, etc. So you would call the generate_response() from `generate_success_response()`, etc.

backend/tests/README.md
```diff
@@ -0,0 +1,5 @@
+`cd /home/prakash/fast-music-remover`
+
+`PYTHONPATH=$(pwd) python -m unittest discover -s backend/tests -p 'test_*.py`
+
+Run the above two commands to test the code
```
omeryusufyagci, line 5: we should use pytest here, also as a side note, usually you would do:

```sh
cd /home/`whoami`/...
``` 

which would work for anyone using a Unix-based OS. 

Let's go with pytest regardless

templates/index.html
```diff
@@ -16,8 +16,8 @@
             })
             .then(response => response.json())
             .then(data => {
-                if (data.status === "completed") {
-                    document.getElementById("videoPlayer").src = data.video_url;
+                if (data.status === "success") {
+                    document.getElementById("videoPlayer").src = data.data.video_url;
```
omeryusufyagci, line 20: data.data typo?

config.json
```diff
@@ -1,7 +1,7 @@
 {
     "deep_filter_path": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
     "deep_filter_tarball_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx.tar.gz",
-    "deep_filter_encoder_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/enc.onnx",
```
omeryusufyagci, line 4: typo

app.py
```diff
@@ -1,193 +1,96 @@
...
         return send_from_directory(directory=app.config["UPLOAD_FOLDER"], path=filename, mimetype="video/mp4")
+    
     except Exception as e:
         logging.error(f"Error serving video: {e}")
-        return jsonify({"status": "error", "message": "Failed to serve video."}), 500
-
+        return ResponseHandler.error("Failed to serve video.", 500)
 
 if __name__ == "__main__":
-    app.run(port=8080, debug=True)
```
omeryusufyagci, line 193: why this port change?

MediaProcessor/src/main.cpp
```diff
@@ -58,4 +58,4 @@ int main(int argc, char* argv[]) {
 
     std::cout << "Video processed successfully: " << processedMediaPath << std::endl;
     return 0;
-}
```
omeryusufyagci, line 61: please run the clang (with the file available on project root, i.e. .clang-format) for MediaProcessor/src or setup the pre-commit hooks which would do this for you

backend/tests/test_media_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        
+        # Set the return value when instantiating YoutubeDL
+        mock_yt_dl_context_manager = MagicMock()
+        mock_yt_dl_context_manager.__enter__.return_value = mock_ydl_instance
+        mock_yt_dlp.return_value = mock_yt_dl_context_manager
+
+        # Test download_media method
+        result = MediaHandler.download_media(self.video_url, self.base_directory)
+        expected_file = Path(self.base_directory) / "Test_Video.mp4"
+        self.assertEqual(result, str(expected_file.resolve()))
```
omeryusufyagci, line 39: Sometimes due to rate limiting yt-dlp takes minutes to finish. Please add a timeout, e.g. 10 seconds, here for getting the info which should be rather fast irrespective of media size.

backend/tests/test_media_handler.py
```diff
@@ -0,0 +1,93 @@
...
+            stdout="Video processed successfully: /path/to/processed_video.mp4",
+            stderr=""
+            
+        )
+        # Test process_with_media_processor
+        result = MediaHandler.process_with_media_processor(
+            self.video_path, self.base_directory, self.config_path
+        )
+
+        self.assertEqual(result, "/path/to/processed_video.mp4")
```
omeryusufyagci, line 55: What's happening here with the assertion on l.55?

omeryusufyagci: You can see [here](https://github.com/omeryusufyagci/fast-music-remover/blob/32352d65b3a1dc30a3c8fbb1ea01af4241b8012f/MediaProcessor/tests/AudioProcessorTester.cpp#L47) for an example on how a similar test was made for the core. You could reuse the same utility, just need to convert it to python code.

backend/tests/test_media_handler.py
```diff
@@ -0,0 +1,93 @@
...
+from unittest.mock import patch, MagicMock
+from pathlib import Path
+from backend.media_handler import MediaHandler
+
+class TestMediaHandler(unittest.TestCase):
+
+    def setUp(self):
+        """Set up base directory and mock paths for testing."""
+        self.base_directory = tempfile.mkdtemp()
+        self.video_url = "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
```
omeryusufyagci, line 12: Please don't place data inside of the code. That later would require code updates for simple test data changes. These should be placed in appropriate test data files. You can see what has been done for the MediaProcessor for examples.

backend/tests/test_media_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        self.config_path = "/path/to/config.json"
+        self.video_path = "/path/to/video.mp4"
```
omeryusufyagci, line 14: Not sure what these mean?

omeryusufyagci (changes requested):
Hey @prakash-2002-ramanathan, thanks for the changes. I took a look and there are a few things I'd like you to address. Could you please take a look? Thanks a lot!

templates/index.html
prakash-2002-ramanathan: No @omeryusufyagci , in the response object that is being returned, the video_url is wrapped inside another data object. Do you have any suggestions on how to improve this ?

backend/tests/test_media_handler.py
prakash-2002-ramanathan: Hey do you want to add the time out on the main code too ?

omeryusufyagci: Yes would be good

backend/tests/test_media_handler.py
omeryusufyagci: Well, `/path/to/...` has to be a real path first. Regarding what we did in the core: we have a known truth, i.e. a processed video that we manually verified by watching it as a reference of what proper filtering is doing to a known source. We then do a byte by byte comparison with that known truth to see if functionality is in tact.

Byte-by-byte is definitely not the best way to do this, but it was a fast solution that actually works with a good tolerance. To be improved, but better than nothing

templates/index.html
omeryusufyagci: It's also because data is used both in the js and the python side. You could rename the data field of the json object to `payload` which is quite std to use.

prakash-2002-ramanathan:
Hey @omeryusufyagci , could you explain how do you want this 
`download_media` function to be tested. Is it ok for you if I mock the  yt_dlp.YoutubeDL() function? I tried to download the youtube video and check it with a previously downloaded video. But sometimes the byte by byte comparison causes problem., since the content on youtube is dynamic and changes everytime when it's downloaded

omeryusufyagci:
> Hey @omeryusufyagci , could you explain how do you want this `download_media` function to be tested. Is it ok for you if I mock the `yt_dlp.YoutubeDL()` function? I tried to download the youtube video and check it with a previously downloaded video. But sometimes the byte by byte comparison causes problem., since the content on youtube is dynamic and changes everytime when it's downloaded

Hey @prakash-2002-ramanathan, you could try to test via the extracted metadata instead of the video itself to test if yt-dlp works fine. If we can get the metadata, we should likely be OK. You could, however, mock it as well. Byte-by-byte is not the best, I know... I'll open an issue about that. Thanks!

prakash-2002-ramanathan:
Hello @omeryusufyagci, I have addressed all your requested changes. Please have a look

backend/tests/test_utils.py
```diff
@@ -0,0 +1,85 @@
...
+    valid_processed_string = Utils.get_processed_video_path(
+        ["Video processed successfully: valid_file-name"]
+    )
+    assert valid_processed_string == "valid_file-name"
+
+    # Test with an invalid string that does not contain the success message
+    invalid_processed_string = Utils.get_processed_video_path(
+        ["Video processed successful"]
+    )
+    assert invalid_processed_string is None
```
omeryusufyagci, line 85: As this is an important test, could you please add better coverage against edge cases. Empty list, strangely formatted messages, etc.

backend/tests/test_media_handler.py
```diff
@@ -0,0 +1,57 @@
+import pytest
+import os
+from concurrent.futures import ThreadPoolExecutor, TimeoutError as FuturesTimeoutError
+from pathlib import Path
+from backend.media_handler import MediaHandler
+import utils as utils
```
omeryusufyagci, line 6: intentional?

backend/tests/test_media_handler.py
```diff
@@ -0,0 +1,57 @@
...
+    assert "duration" in metadata, "Duration missing in metadata"
+    assert metadata["duration"] > 0, "Invalid duration in metadata"
+
+
+
+def test_process_with_media_processor():
+    """Test video processing with MediaHandler."""
+    BASE_DIR = Path(__file__).resolve().parents[2]
+    TEST_DIR = Path(__file__).parent.resolve()
+    DEEPFILTERNET_PATH = str((BASE_DIR / utils.DEEP_FILTER_PATH).resolve())
```
omeryusufyagci, line 36: would be good to have a `pytest.fixture` to handle these.

omeryusufyagci (changes requested):
Hi @prakash-2002-ramanathan, thank you for the updates. 

I've also updated branch `34-...` to reduce the amount of clutter as it was getting hard to review.

I noticed the changes for core are still present in this PR. Could you please rebase your commit history to remove them from here? We should focus on those on a separate PR after this is finished. 

Regarding the updates, in addition to the comments I left, here are some more global remarks:
- Please try to ensure the test names reflect exactly what we're testing for and what do we expect in return. 
- Try to utilize more the features of the testing framework such as fixtures, and parameters. This is to make it easier to extend these tests later. 

I appreciate your effort on this and looking forward to merge it soon! In the meantime, have you noticed we have a [discord](https://discord.com/invite/xje3PQTEYp) now? If you have any questions, or just want to say hello, pass by!

omeryusufyagci:
@prakash-2002-ramanathan please note the update on the branch you're targeting that was required to add urgent functionalities. Please ensure the added functionality is kept in this PR. Thank you!

Issue #55, opened 2024-10-23
Bug: Cross platform compilation support is broken for core


omeryusufyagci:
`MediaProcessor::AudioProcessor::invokeDeepFilterNetFFI()` breaks cross platform support as it relies on a .so that was compiled with `x86_64-unknown-linux-gnu` flag. 

We have to compile the rust library for `x86_64-pc-windows-gnu`, `x86_64-apple-darwin`, `aarch64-apple-darwin`, `x86_64-unknown-linux-gnu`, and update the top level CMake file to ensure we link the correct one wrt the platform.

```sh
$ make
[ 10%] Linking CXX executable MediaProcessor
ld: warning: dylib (/opt/homebrew/Cellar/libsndfile/1.2.2/lib/libsndfile.dylib) was built for newer macOS version (14.0) than being linked (13.3)
ld: warning: ignoring file /Users/oyagci/fast-music-remover/MediaProcessor/lib/libdf.so, building for macOS-arm64 but attempting to link with file built for unknown-unsupported file format ( 0x7F 0x45 0x4C 0x46 0x02 0x01 0x01 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 )
Undefined symbols for architecture arm64:
  "_df_create", referenced from:
      MediaProcessor::AudioProcessor::invokeDeepFilterFFI(std::__1::__fs::filesystem::path) in AudioProcessor.cpp.o
  "_df_free", referenced from:
      MediaProcessor::AudioProcessor::invokeDeepFilterFFI(std::__1::__fs::filesystem::path) in AudioProcessor.cpp.o
  "_df_get_frame_length", referenced from:
      MediaProcessor::AudioProcessor::invokeDeepFilterFFI(std::__1::__fs::filesystem::path) in AudioProcessor.cpp.o
  "_df_process_frame", referenced from:
      MediaProcessor::AudioProcessor::invokeDeepFilterFFI(std::__1::__fs::filesystem::path) in AudioProcessor.cpp.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [MediaProcessor] Error 1
make[1]: *** [CMakeFiles/MediaProcessor.dir/all] Error 2
make: *** [all] Error 2
```

omeryusufyagci:
[...]

PR #61, opened 2024-10-27
Fix cross platform builds by adding a dedicated workflow

omeryusufyagci:
As reported in #55, cross platform builds broke with the introduction of the linux .so as a dependency. 
This PR introduces a new workflow that will take care of building the shared objects for each supported platform.


omeryusufyagci:
A few local tests showed some promise, but job times hitting 6+ minutes seems rather inefficient. 
Main problem is to provide x86_64 MacOS builds for the C API of the filter. Otherwise we have support for Windows, Linux and arm based new MacOS machines.

Issue #60, opened 2024-10-26
Feature: Add `FFmpegController` to core


omeryusufyagci:
Implement the FFmpegController which will serve as the high level orchestrator for FFmpeg operations. For more context, see #45.

One thing to note here is that, the user may wish to update the FFmepg settings after instantiating the controller. We have to provide an `updateSettings()` public member function to faciliate this. 

Initially I see 2 feasible solutions which would require changes to the `FFmepgSettingsManager`:

1. Add a private member varible of type unordered map to keep track of modification status of each setting. Get the map within the controller, check for changed fields when `updateSettings()` is called.
2. Same thing as (1), but with a bitmask. Introduce a public enum, e.g. `SettingsFlags` to expose the mask of each field, e.g.`OverwriteFlag = 0x01`. 

With the limited amount of settings and likely infrequent nature of these calls, hash map seems like the better option for readability.

omeryusufyagci:
Here are the current use cases we have:
- [`extractAudio()`](https://github.com/omeryusufyagci/fast-music-remover/blob/a4baf434bf34c0a3e47bb1494a387b44d9fe5aa7/MediaProcessor/src/AudioProcessor.cpp#L76)

- [`generateChunkFile()`](https://github.com/omeryusufyagci/fast-music-remover/blob/a4baf434bf34c0a3e47bb1494a387b44d9fe5aa7/MediaProcessor/src/AudioProcessor.cpp#L118)

- [`mergeChunks()`](https://github.com/omeryusufyagci/fast-music-remover/blob/a4baf434bf34c0a3e47bb1494a387b44d9fe5aa7/MediaProcessor/src/AudioProcessor.cpp#L295)

- [`mergeMedia()`](https://github.com/omeryusufyagci/fast-music-remover/blob/a4baf434bf34c0a3e47bb1494a387b44d9fe5aa7/MediaProcessor/src/VideoProcessor.cpp#L20)

The initial implementation of the controller should support these operations with a possibility of updating the settings as mentioned in the description.

PR #63, opened 2024-10-29
Feature: Add ffmpegcontroller to core (#60)

andcscott:
Hi @omeryusufyagci 

I may have jumped the gun on this, but I'm eager for your feedback and I figure we'll go through a couple rounds of iteration to make sure it's right anyway.

I have three commits for you this time:

1. Update `FFmpegCommandBuilder` to prepare for `FFmpegController`
2. Update `FFmpegConfigManager` for the same purpose
3. Finally add `FFmpegController` :1st_place_medal: 

I've attempted to be specific with the changes made in each commit message, please let me know if there are any concerns.

Still todo (not added yet as I assume they'd be revised based on any required changes):

- [ ] Test suite
- [ ] Docstrings for most new methods
- [ ] `updateSettings()` (I really like the idea of a bit field, but you're probably correct than an enum is the way to go for now)


MediaProcessor/src/FFmpegCommandBuilder.h
```diff
@@ -31,70 +39,115 @@ class FFmpegCommandBuilder : public CommandBuilder {
     FFmpegCommandBuilder& addOverwrite();
 
     /**
-     * @brief Adds input flag and input file path
+     * @brief Adds input flag and the input file path from FFmpegSettingsManager
```
omeryusufyagci, line 42: Perhaps a find-replace for FFmpegSettingsManager as it's now the FFmpegConfigManager

MediaProcessor/src/FFmpegCommandBuilder.cpp
```diff
@@ -6,71 +6,139 @@
...
+
+FFmpegCommandBuilder& FFmpegCommandBuilder::addFilterComplex(
+    const std::vector<fs::path> processedChunkCol) {
+    addFlag("-filter_complex", buildFilterComplex(processedChunkCol));
+
+    return *this;
+}
+
+std::string FFmpegCommandBuilder::buildCrossfade(int chunkColSize) const {
+    std::string filterComplex = "";
```
omeryusufyagci, line 112: not that important but let's remove the `= ""` as the default string ctor does that for us. Also, good call on moving this guy here.

omeryusufyagci (changes requested):
Hi @andcscott, thanks a lot, this is a solid start! 

This provides a good initial implementation and it's flexible enough that we can tweak it along the way if we need to. 

Of course ensuring there is no regression of functionality is the most important part, and we can take a closer look at that with your tests. 

Thanks again!

P.S. sorry I couldn't look at this sooner, it was quite a busy week.

andcscott:
HI @omeryusufyagci just wanted to let you know I saw your comments, will be working on it this week, sorry things got a little crazy here too!

MediaProcessor/src/FFmpegCommandBuilder.h
```diff
@@ -31,70 +39,115 @@ class FFmpegCommandBuilder : public CommandBuilder {
...
+    FFmpegCommandBuilder& addFilterComplex(const std::vector<fs::path> processedChunkCol);
+
     /**
      * @brief Constructs the command string from the added arguments and flags
      *
      * @return The FFmpeg command as a string
-     *
-     * @throws std::runtime_error If input or output file path is missing
      */
     std::string build() const override;
```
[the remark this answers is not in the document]
andcscott: Nope, this was left over from a prior change - removed!

MediaProcessor/src/FFmpegCommandBuilder.cpp
```diff
@@ -6,71 +6,139 @@
...
 
-std::string FFmpegCommandBuilder::build() const {
-    if (m_inputFilePath.empty()) {
-        throw std::runtime_error("Input file path must be specified.");
-    }
+FFmpegCommandBuilder& FFmpegCommandBuilder::addStrictExperimental() {
+    addFlag("-strict", "experimental");
+
+    return *this;
+}
```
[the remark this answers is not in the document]
andcscott: Added additional options similar to the codec enums

MediaProcessor/src/FFmpegCommandBuilder.cpp
```diff
@@ -6,71 +6,139 @@
 
 namespace MediaProcessor {
 
-FFmpegCommandBuilder::FFmpegCommandBuilder(FFmpegConfigManager& ffmpegConfig)
+FFmpegCommandBuilder::FFmpegCommandBuilder(const FFmpegConfigManager& ffmpegConfig)
     : m_ffmpegConfig(ffmpegConfig), m_ffmpegPath(ConfigManager::getInstance().getFFmpegPath()) {}
```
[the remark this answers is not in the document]
andcscott: You're right, for some reason I was doing this in the ctor for the controller instead. Fixed.

MediaProcessor/src/FFmpegCommandBuilder.h
andcscott: Good eyes, that was the last remaining reference to the old settings manager

andcscott:
In addition to the requested changes I also took a pass at adding a test for the public API. Nothing squashed yet to help separate the changes from the additions. Thanks!

MediaProcessor/tests/FFmpegControllerTester.cpp
```diff
@@ -0,0 +1,65 @@
...
+    configManager.setAudioChannels(1);
+    configManager.setAudioCodec(AudioCodec::AAC);
+    configManager.setOutputFilePath(outputPath);
+
+    FFmpegController controller = FFmpegController(configManager);
+
+    EXPECT_EQ(controller.extractAudio(), true);
+
+    EXPECT_TRUE(fs::exists(outputPath));
+}
```
omeryusufyagci, line 63: Could you take a look at [this utility](https://github.com/omeryusufyagci/fast-music-remover/blob/8c4952ea58d8c195c8ba08ede40252cc7359b67d/MediaProcessor/tests/AudioProcessorTester.cpp#L59) to see if you can reuse it here? It's more an integration test than a unit test per se, but critical to ensure we don't introduce filtering performance issues as we do these big changes.

omeryusufyagci (commented):
Overall, these look pretty good @andcscott, thanks a bunch! 

Would be great if we could add the test I mentioned below. Then, `updateSettings()` is up next to conclude this I believe?

andcscott:
Hi @omeryusufyagci, I don't know about you, but it feels pretty good to have this section (almost) ready!

1. Added the requested utility to the tests. I don't have a ton of experience with gtest. It seemed straightforward, but of course let me know if there's a problem.
2. Added `updateSettings`. This implementation uses bitmasks as suggested in #60. While the hash map would have likely been less tedious to implement, I'm hopeful the added efficiency helps "future proof" the class.
3. Added docstrings.

Nothing squashed or merged yet, please let me know how you want to proceed with that.

Thanks!

MediaProcessor/src/FFmpegConfigManager.cpp
```diff
@@ -77,12 +139,106 @@ VideoCodec FFmpegConfigManager::getVideoCodec() const {
...
+
+    if (globalBitField & GlobalConfigMask::InputFilePath) {
+        m_globalSettings.inputFilePath = inputFilePath;
+        result |= GlobalConfigMask::InputFilePath;
+    }
+
+    if (globalBitField & GlobalConfigMask::OutputFilePath) {
+        m_globalSettings.outputFilePath = outputFilePath;
+        result |= GlobalConfigMask::OutputFilePath;
+    }
```
omeryusufyagci, line 188: I know it was me who proposed the idea of this "check for changed fields", but I feel it's difficult to justify it. We're going to add all this code just so we can update only the changed settings... If we were to make a full update like a `PUT`, it'd be so simple:

```cpp
m_globalSettings = globalValues;  // just assign the new struct
```
and we'd be done. 

I've spent some time to see how we can avoid passing a bunch of args and to avoid cascaded contional checks. 
We could do it with a new helper class and some template functions, but at that stage it really feels like overkill for a relatively cheap operation. 

I really appreciate you taking this on, but I think going for a full update here would keep things simple and maintainable, but please let me know what you think!

MediaProcessor/src/FFmpegConfigManager.h
```diff
@@ -49,23 +94,40 @@ class FFmpegConfigManager {
...
+    const std::unordered_map<VideoCodec, std::string>& getVideoCodecAsString() const;
+    const std::unordered_map<CodecStrictness, std::string>& getCodecStrictnessAsString() const;
+
+    // Update Global, Audio, or Video Settings
+    int updateSettings(int globalBitField, int audioBitField, int videoBitField, bool overwrite,
+                       CodecStrictness strictness, const fs::path& inputFilePath,
+                       const fs::path& outputFilePath, AudioCodec codec, const fs::path& chunksPath,
+                       const std::vector<fs::path>& processedChunkCol, int chunkIndex,
+                       int audioSampleRate, int audioChannels, double audioStartTime,
+                       double audioDuration, double overlapDuration, VideoCodec videoCodec);
```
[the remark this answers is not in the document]
omeryusufyagci: Please see the comment on the implementation as it voids this mostly.

MediaProcessor/tests/FFmpegControllerTester.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    FFmpegController controller = FFmpegController(configManager);
+
+    EXPECT_EQ(controller.extractAudio(), true);
+
+    EXPECT_TRUE(fs::exists(testAudioOutputPath));
+
+    EXPECT_TRUE(
+        TestUtils::CompareFiles::compareAudioFiles(testAudioOutputPath, testAudioProcessedPath));
+}
```
omeryusufyagci, line 67: I haven't run this yet, but looks reasonable. Could you also add tests for the remaining public member fns and I'll run the tests on my end as well.

omeryusufyagci (commented):
Hi @andcscott, thanks a lot for all this! Indeed, looking forward to merge these! 

In fairness, my idea of adding either a hashmap or bitmask for only updating the changed settigns did not pan out to be a good one. It's difficult to justify the cost -depending on the implementation, either added complexity or repetitives, for a relatively cheap operation. 

I tried to explain a bit better in-place and please let me know what you think. I tend to side on going with the easier option of doing a full update, adding the remaining tests and merging this sooner than later. 

As always, I'm open to discuss any parts of it. On the topic, we now have a [discord](https://discord.com/invite/xje3PQTEYp) if you prefer that. 

Thanks again and wish you a nice rest of your weekend!

MediaProcessor/src/FFmpegConfigManager.cpp
andcscott: Hi @omeryusufyagci, no worries, it is definitely a lot of extra code! I have a few thoughts on how do do this, and I'm curious what you prefer. I suppose it depends on how "full" a full update should be.

1. Keep it simple with overloads, one version of updateSettings() for each FFmpeg settings struct
2. Use std::optional so the user can pass 1-2 structs without overwriting the third
3. Require a **full** update, user must pass 3 new structs

If you had something else in mind let me know, that's just how I'm reading it right now. Thanks for the discord link too! I don't spend a lot of time on there but I'll join the channel :)

MediaProcessor/tests/FFmpegControllerTester.cpp
andcscott: Hi @omeryusufyagci ...just so you know I'm working on this ;)

I needed to do some refactoring. When implementing the controller I added generateChunkFile as a public method, but that presents problems for the tests because the helpers it are private to AudioController. To me it seems easier for the user if we add splitAudioIntoChunks as public to the controller, with helpers like generateChunkFile as private.

Short version: I'm making my life harder in the short term by moving more pieces out of AudioController so we can hopefully have an easier time moving forward with tests and setting up a nice API :)

omeryusufyagci: Thanks @andcscott, that sounds good! 

Some ffmpeg commands work both for audio and video, perhaps this works the same as well. If that's the case, we can name it like split**Media**...and maybe a name like `chunkMedia()` will read easier for the public fn as well.

andcscott:
Hi @omeryusufyagci, first of all sorry this took me so long, these past couple weeks have been crazy! In addition to the requested changes I also cleaned up the commits a little bit, it was getting a little hard to keep track (at least on my end). This is a holiday weekend for us in the States so I at least wanted to make sure I pushed something before that. As always, please let me know if any additional changes are required. Thanks!

omeryusufyagci:
> Hi @omeryusufyagci, first of all sorry this took me so long, these past couple weeks have been crazy! In addition to the requested changes I also cleaned up the commits a little bit, it was getting a little hard to keep track (at least on my end). This is a holiday weekend for us in the States so I at least wanted to make sure I pushed something before that. As always, please let me know if any additional changes are required. Thanks!

Hi @andcscott, thanks a lot for the updates, and absolutely no worries on the schedule! I really appreciate your work on the FFmpeg API, and so glad to see how it evolved and that we're almost ready to merge it all! 

Overall, it's pretty much there; my only consideration is as I commented in-place, the somewhat unclear separation of responsibilities between AudioProcessor and FFmpegConfigManager. I'd be interested to hear your thoughts on this as well, but personally, I'd feel better if FFmpegConfigManager didn't store non-FFmpeg related data, like overlap durations or chunk sizes. 
The rest, such as codecs, etc. should remain in FFmpegConfigManager, but it'd be better if we received the audio-specific information from the AudioProcessor.

MediaProcessor/src/FFmpegConfigManager.h
```diff
@@ -20,27 +25,56 @@ enum class VideoCodec { H264, H265, VP8, VP9, UNKNOWN };
...
+        int numChunks;
+        int sampleRate = 48000;
+        int numChannels = 2;
+        double overlapDuration = DEFAULT_OVERLAP_DURATION;
+    } m_audioSettings;
+
+    struct FFmpegVideoSettings {
+        VideoCodec codec = VideoCodec::H264;
+    } m_videoSettings;
+
```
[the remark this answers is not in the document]
andcscott: I think I share your opinion on the separation between the two classes. I had only relocated the variables for ease of access (especially for the tests), but I do agree they're not really "config" options. 

I'll get rid of `int numChunks` and `double overlapDuration`, what do you think about `fs::path chunksPath` and `std::vector<fs::path> processedChunkCol`? Personally I see the chunks path as redundant (that is, I should remove it), while the other arguably belongs to both **but** maybe more on the side of AudioProcessor?

I'm also curious how you feel about my moving methods away from AudioProcessor to the controller. If you feel like they are better suited to AudioProcessor I have no problems moving them back.

Re: the structs - it was intentional - with updateSettings taking the structs rather than individual options won't we have to make them public so the user can initialize them? Sorry if I'm missing something.

andcscott:
Hey @omeryusufyagci, I'm glad the schedule is no issue!

I'm inclined to agree with you on the vars for the two classes, no reason to muddy things when we already have clearly defined roles. I replied to your in-place comment with more details, but basically I'm looking to see 1) what other vars you think should be moved back to AudioProcessor (if any), 2) your thoughts on moving AudioProcessor's methods since I'm moving some vars back, and 3) alternatives to making the structs public. Let me know what you think, thanks!

MediaProcessor/src/FFmpegConfigManager.h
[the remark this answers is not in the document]
omeryusufyagci: > I'll get rid of int numChunks and double overlapDuration, what do you think about fs::path chunksPath and std::vector<fs::path> processedChunkCol? Personally I see the chunks path as redundant (that is, I should remove it), while the other arguably belongs to both but maybe more on the side of AudioProcessor?

I think the latter is also something the AudioProcessor should own and manage. Thinking from the FFmpeg side, we shouldn't necessarily be concerned with if the input is in chunks or a full file, etc. Those are AudioProcessor's duties. This also would eliminate the redundancy as you said. I do see it will make testing a little less convenient, but in terms of design I feel it would be more clear. So, instead of holding copies of strictly audio-related information, we should receive them where relevant as arguments. Let me know if you see something not feasible with this though.  

> I'm also curious how you feel about my moving methods away from AudioProcessor to the controller. If you feel like they are better suited to AudioProcessor I have no problems moving them back.

Any functionality that uses FFmpeg should reside in the FFmpegController. So, I think that was a good decision.

> Re: the structs - it was intentional - with updateSettings taking the structs rather than individual options won't we have to make them public so the user can initialize them? Sorry if I'm missing something.

You're right, sure. I'm still keen on keeping the structs private so for now we can make the controller a friend as the two are instrinsically coupled anyway. 

Thank you and please let me know if I can help in any way!

andcscott:
Hi @omeryusufyagci, the latest push addresses the issues of separation between AudioProcessor and FFmpeg classes and updates the corresponding tests. Thanks!

omeryusufyagci (approved):
Hi @andcscott, thank you for the changes! These cover what we've discussed and I believe serves as a good initial implementation for our FFmpeg API. 

The dedicated branch was stale and I've updated it. Could you please rebase your branch and let me know, and I'll accept this PR? Thanks!

andcscott:
Hi @omeryusufyagci just rebased and pushed, let me know if I missed anything, thanks!

PR #68, opened 2024-11-11
Updated pyproject.toml: configured black to use line-length = 120

nikhiljangra264:
A little Improvement. Curretly we have to manualy put 
```bash
black . --line-length 120
```
before each command
but now we can do
```bash
black .
```


omeryusufyagci (approved):
Sure, thanks!

PR #75, opened 2024-11-19
Feature: Add a project launcher #67

### LINKED ISSUE #67: Feature: Add a project launcher
This will manage the platform specific configuration/installation of dependencies. It should automatically update config.json file so that we can avoid adding a separate field for each platform and dependency that the user has to verify and maintain. 

This will also introduce more flexibility in how start our applications built on top of MediaProcessor.  
The launcher should not assume the current web-based application will be the only use case. In the future we might provide a desktop GUI, and potentially other standalone applications. 

The goal is to achieve a cross-platform seamless user experience.

nikhiljangra264:
closes #67 

Added launcher.py
1. launch web application
```python
python launcher.py --mode=web
```
2. install dependencies
```python
python launcher.py --install
```
3. rebuilding backend MediaProcessor
```python
python launcher.py --rebuild
```
4. help
```python
python launcher.py --help
```

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -1,0 +1,479 @@
+import argparse
+import atexit
+import json
+import logging
+import logging.config
+import os
+import platform
+import subprocess
+import sys
+import threading
+import time
+import webbrowser
+from pathlib import Path
+
+CONFIG_FILE = os.path.join("config.json")
+venv_name = "virtual_env"  # For downloading python packages
+venv_dir = Path.cwd() / venv_name  # Generate the path for the virtual environment
+DEFAULT_CONFIG = {
+    "version": 1,
+    "formatters": {
+        "detailed": {"format": "%(asctime)s [%(levelname)s] [%(funcName)s(), %(lineno)d]: %(message)s"},
+        "simple": {"format": "[%(levelname)s] %(message)s"},
+    },
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+            "formatter": "simple",
+            "stream": "ext://sys.stdout",
+        },
+    },
+    "root": {"level": "ERROR", "handlers": ["console"]},
+}
+
+
+def setup_logging(log_level, log_file=False):
+    DEFAULT_CONFIG["root"]["level"] = log_level
+    if log_file:
+        DEFAULT_CONFIG["handlers"]["file"] = {
+            "class": "logging.handlers.RotatingFileHandler",
+            "formatter": "detailed",
+            "filename": f"logfile_{time.strftime('%Y%m%d_%H%M%S')}.log",
+            "maxBytes": 1024 * 1024 * 5,  # 5MB
+            "backupCount": 3,
+        }
+        DEFAULT_CONFIG["root"]["handlers"].append("file")
+
+    logging.config.dictConfig(DEFAULT_CONFIG)
+
+
+def run_command(command, cwd=None):
+    """
+    Executes a shell command.
+
+    Args:
+        command (str): The command to run.
+
+    Raises:
+        subprocess.CalledProcessError: If any command fails.
+    """
+    logging.debug(f"Executing command: {command}")
+    process = subprocess.Popen(
+        command.split(),
+        cwd=cwd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    stdout, stderr = process.communicate()  # wait for process to terminate
+    if stdout:
+        logging.debug(f"Command output: {stdout}")
+    if stderr:
+        logging.error(f"Command error: {stderr}")
+
+    if process.returncode != 0:
+        raise subprocess.CalledProcessError(process.returncode, command)
+
+
+class DependencyHandler:
+    """
+    Represents a dependency with methods for checking and installing it.
+
+    Attributes:
+        name (str): The name of the dependency.
+        package_name (dict): { platform: package_name }.
+        check_cmd (dict): { platform: list of command or function to check if the dependency is installed }.
+        install_cmd (dict): { platform: list of command or function to install the dependency }.
+
+        if package_name is not provided, it will be set to the name.
+
+        if check_cmd is not provided
+            package_name --version will be used to check if the dependency is installed
+
+        if install cmd is not provided
+            for linux: package manager based on distro
+            for macos: brew
+            for windows: msys2
+    """
+
+    def __init__(self, name, package_name=None, check_cmd=None, install_cmd=None):
+        self.name = name
+        self.package_name = package_name or {}
+        self.check_cmd = check_cmd or {}
+        self.install_cmd = install_cmd or {}
+        self._installers = {
+            "Linux": self._get_install_commands_linux,
+            "Darwin": self._get_install_commands_darwin,
+            "Windows": self._get_install_commands_windows,
+        }
+        # Mapping of Linux distributions to package managers
+        self._linux_distro_map = {
+            "ubuntu": "sudo apt-get install -y",
+            "debian": "sudo apt-get install -y",
+            "kali": "sudo apt-get install -y ",
+            "pop": "sudo apt-get install -y ",
+            "elementary": "sudo apt-get install -y ",
+            "mint": "sudo apt-get install -y",
+            "fedora": "sudo dnf install -y",
+            "rhel": "sudo dnf install -y",
+            "centos": "sudo dnf install -y",
+            "rocky": "sudo dnf install -y",
+            "alma": "sudo dnf install -y",
+            "arch": "sudo pacman -S --needed --noconfirm",
+            "manjaro": "sudo pacman -S --needed --noconfirm",
+            "endeavouros": "sudo pacman -S --needed --noconfirm",
+            "garuda": "sudo pacman -S --needed --noconfirm",
+            "opensuse": "sudo zypper install -y",
+            "suse": "sudo zypper install -y",
+            "alpine": "sudo apk add",
+            "solus": "sudo eopkg install -y",
+            "void": "sudo xbps-install -y",
+            "clearlinux": "sudo swupd bundle-add",
+        }
+
+    def ensure_dependency_installed(self, system, install=True):
+        """
+        Checks if the dependency is installed. If not installs it if install arg is True.
+        """
+        commands = self._get_check_commands(system)
+        try:
+            for cmd in commands:
+                if type(cmd) == str:
+                    run_command(cmd)
+                else:
+                    cmd()
+
+        except (FileNotFoundError, subprocess.CalledProcessError):
+            if not install:
+                logging.info(f"{self.name} is not installed.")
+                sys.exit(1)
+            self.install_dependency(system)
+
+        except Exception as e:
+            logging.error(f"While Checking Dependency: {e}", exc_info=True)
+            sys.exit(1)
+
+    def install_dependency(self, system):
+        """
+        Installs the dependency for the specified operating system.
+
+        Args:
+            system (str): The operating system type.
+        """
+        try:
+            commands = self._get_install_commands(system)
+            for cmd in commands:
+                if type(cmd) == str:
+                    run_command(cmd)
+                else:
+                    cmd()
+        except Exception as e:
+            logging.error(f"While Installing: {e}.", exc_info=True)
+            sys.exit(1)
+
+    def _get_check_commands(self, system):
+        if system in self.check_cmd:
+            return self.check_cmd[system]
+        if "all" in self.check_cmd:
+            return self.check_cmd["all"]
+        return [f"{self.package_name.get(system, self.name)} --version"]
+
+    def _get_install_commands(self, system):
+        if system in self.install_cmd:
+            return self.install_cmd[system]
+        if "all" in self.install_cmd:
+            return self.install_cmd["all"]
+        if system not in self._installers:
+            logging.error(f"Unspported sytem {system}.")
+            sys.exit(1)
+
+        return self._installers[system]()
+
+    def _get_install_commands_linux(self):
+        import distro
+
+        distro = distro.id().lower()
+        if distro not in self._linux_distro_map:
+            logging.error(f"Unsupported linux distro {distro}.")
+            sys.exit(1)
+
+        return [f"{self._linux_distro_map[distro]} {self.package_name.get('Linux', self.name)}"]
+
+    def _get_install_commands_darwin(self):
+        return [f"brew install {self.package_name.get('Darwin', self.name)}"]
+
+    def _get_install_commands_windows(self):
+        if not check_msys2_installed():
+            install_msys2()
+        return [f"pacman -S --needed --noconfirm {self.package_name.get('Windows', self.name)}"]
+
+
+# pkg-config should come before sndfie and nlohmann-json
+dependencies = [
+    DependencyHandler("cmake", {"Windows": "mingw-w64-x86_64-cmake"}, {"all": ["cmake --version"]}),
+    DependencyHandler(
+        "g++",
+        {"Windows": "mingw-w64-x86_64-gcc", "Darwin": "gcc"},
+        {"all": ["g++ --version"]},
+        {"Windows": ["pacman -S --needed --noconfirm base-devel mingw-w64-x86_64-toolchain"]},
+    ),
+    DependencyHandler("pkg-config"),
+    DependencyHandler("ffmpeg", {"Windows": "mingw-w64-x86_64-ffmpeg"}, {"all": ["ffmpeg -version"]}),
+    DependencyHandler(
+        "libsndfile",
+        {"Windows": "mingw-w64-x86_64-libsndfile", "Linux": "libsndfile1-dev"},
+        {"all": ["pkg-config --exists sndfile"]},
+    ),
+    DependencyHandler(
+        "nlohmann-json",
+        {"Windows": "mingw-w64-x86_64-nlohmann-json", "Linux": "nlohmann-json3-dev"},
+        {"all": ["pkg-config --exists nlohmann_json"]},
+    ),
+]
+
+
+def check_msys2_installed():
+    try:
+        run_command("pacman --version")
+        return True
+    except (FileNotFoundError, subprocess.CalledProcessError):
+        return False
+
+
+def check_internet_connectivity():
+    """For Windows check internet connectivity"""
+    try:
+        logging.debug("Checking internet connectivity... ")
+        # Ping Google's public DNS server
+        run_command("ping -n 1 8.8.8.8")
+        logging.debug("Internet connectivity OK")
+    except subprocess.CalledProcessError:
+        logging.error("No internet connection detected.")
+        sys.exit(1)
+
+
+def install_msys2():
+    try:
+        installer_url = (
+            "https://github.com/msys2/msys2-installer/releases/download/nightly-x86_64/msys2-base-x86_64-latest.sfx.exe"
+        )
+        installer_name = "msys2-installer.exe"
+
+        msys2_root_path = "C:\\msys64"
+
+        check_internet_connectivity()
+
+        logging.info("Downloading MSYS2 installer...")
+        logging.debug(f"Installer URL: {installer_url}")
+        run_command(f"curl -L -o {installer_name} {installer_url}")
+
+        logging.info("Running MSYS2 installer...")
+        logging.debug(f"Installing MSYS2 at {msys2_root_path}")
+        run_command(f"{installer_name} -y -oC:\\")
+
+        logging.info("Updating MSYS2 packages...")
+        run_command(f"{msys2_root_path}\\usr\\bin\\bash.exe -lc 'pacman -Syu --noconfirm'")
+
+        logging.info("Editing Environment Variables...")
+        logging.debug(
+            f"Adding {msys2_root_path}\\usr\\bin, {msys2_root_path}\\mingw64\\bin, {msys2_root_path}\\mingw32\\bin to PATH"
+        )
+        # Set it permanently for the current user
+        commands = f"""
+        $oldPath = [Environment]::GetEnvironmentVariable("Path", "User")
+        $newPath = "{msys2_root_path}\\usr\\bin;{msys2_root_path}\\mingw64\\bin;{msys2_root_path}\\mingw32\\bin;" + $oldPath
+        [Environment]::SetEnvironmentVariable("Path", $newPath, "User")
+        """
+        # Run the PowerShell commands
+        subprocess.check_call(["powershell", "-Command", commands])
+
+        # Add MSYS2 paths to the PATH environment variable for the current session
+        logging.debug(
+            f"Addiing {msys2_root_path}\\usr\\bin, {msys2_root_path}\\mingw64\\bin, {msys2_root_path}\\mingw32\\bin to current enviornment."
+        )
+        current_path = os.environ.get("PATH", "")
+        new_path = (
+            f"{msys2_root_path}\\usr\\bin;{msys2_root_path}\\mingw64\\bin;{msys2_root_path}\\mingw32\\bin;"
+            + current_path
+        )
+        os.environ["PATH"] = new_path
+
+        logging.info("MSYS2 installed and updated successfully.")
+        logging.info("NOTE: Please restart your terminal before running this script again.")
+
+    except subprocess.CalledProcessError as e:
+        logging.error(f"Error installing MSYS2: {e}", exc_info=True)
+        sys.exit(1)
+    finally:
+        if os.path.exists(installer_name):
+            os.remove(installer_name)
+
+
+def check_MediaProcessor(system):
+    if system == "Windows":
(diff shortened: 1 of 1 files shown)


launcher.py
```diff
@@ -0,0 +1,358 @@
...
+import time
+import webbrowser
+from pathlib import Path
+
+CONFIG_FILE = os.path.join("config.json")
+venv_name = "virtual_env"  # For downloading python packages
+venv_dir = Path.cwd() / venv_name  # Create the path for the virtual environment
+
+
+def execute_command(command, cwd=None):
```
omeryusufyagci, line 17: For both these lines l14 and l17, as a general remark: let's try to reuse the existing verbs for these actions. For example to make something we've strictly used generate instead of create, and similarly to perform something we've currently used runCommand instead of execute. Please search for similar occurrences in general

launcher.py
```diff
@@ -0,0 +1,358 @@
...
+    Args:
+        commands (str): The command to execute.
+
+    Raises:
+        subprocess.CalledProcessError: If any command fails.
+    """
+    subprocess.check_call(command.split(), cwd=cwd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+
+
+class Dependency:
```
omeryusufyagci, line 30: Could be more descriptive here, e.g DependencyManager as it seems to mainly manage deps?

launcher.py
```diff
@@ -0,0 +1,358 @@
...
+                    sys.exit(1)
+            elif system == "Darwin":
+                install_cmd = [f"brew install {self.package_name.get(system, self.name)}"]
+            elif system == "Windows":
+                if not check_msys2_installed():
+                    install_msys2()
+                install_cmd = [f"pacman -S --needed --noconfirm {self.package_name.get(system, self.name)}"]
+            else:
+                print(f"Unsupported operating system: {system}. Please install {self.name} manually.")
+                sys.exit(1)
```
omeryusufyagci, line 122: This is a little too much nesting and I believe we can do better. How about you introduce member functions for each platform so this reads much lighter?

Perhaps introduce an install() fn that will ask for the appropriate installer from a _get_installer(). The installers could be stored in a hash map. So you'd be injecting this dependency where you're intending to use it. It should eliminate the need for a lot of the control flow here as well. 

Btw, a side note, can you introduce a leading underscore for private member functions?

launcher.py
```diff
@@ -0,0 +1,358 @@
...
+
+        with open(CONFIG_FILE, "w") as config_file:
+            json.dump(config, config_file, indent=4)
+
+    except FileNotFoundError:
+        print(f"{CONFIG_FILE} not found. Please create a default config.json file.")
+        sys.exit(1)
+
+
+def launch_web_application(system):
```
omeryusufyagci, line 297: Would be good to see the logs and even perhaps start it with debug level. Currently if the process fails there is no way to see what happened when started via the launcher.

omeryusufyagci (changes requested):
Hi @nikhiljangra264, thanks a lot for this great start! I've also fixed 2 bugs while testing this version and have already pushed them here. For the rest, I've taken a quick look and left some comments.

MediaProcessor/cmake/src.cmake
```diff
@@ -28,21 +28,39 @@ if(APPLE)
     endif()
 elseif(WIN32)
     set(DF_LIBRARY ${CMAKE_SOURCE_DIR}/lib/libdf.dll.a) # .dll.a for linking, .dll for runtime
+    set(DF_DLL_PATH ${CMAKE_SOURCE_DIR}/lib/df.dll)   # Path for the actual DLL
```
omeryusufyagci, line 31: Why was this needed? Normally providing the .dll.a is enough for linking?

MediaProcessor/cmake/test.cmake
```diff
@@ -14,10 +14,22 @@ if(NOT GTEST_FOUND)
...
+file(TO_CMAKE_PATH "${CMAKE_SOURCE_DIR}/tests/TestMedia" TEST_MEDIA_DIR)
+set(TEST_MEDIA_DIR "${TEST_MEDIA_DIR}" CACHE PATH "Path to test media files")
+
+# Set platform-specific dynamic library for df
+if(APPLE)
+    set(DF_LIBRARY ${CMAKE_SOURCE_DIR}/lib/libdf.dylib)  # macOS dynamic library
+elseif(WIN32)
+    set(DF_LIBRARY ${CMAKE_SOURCE_DIR}/lib/df.dll)  # Windows dynamic library
+elseif(UNIX)
+    set(DF_LIBRARY ${CMAKE_SOURCE_DIR}/lib/libdf.so)  # Linux dynamic library
```
omeryusufyagci, line 26: we could do without the comments at the end of the lines

MediaProcessor/tests/TestUtils.cpp
```diff
@@ -79,12 +79,12 @@ bool CompareFiles::compareFilesByteByByte(const fs::path& filePath1, const fs::p
 bool CompareFiles::compareAudioFiles(const fs::path& filePath1, const fs::path& filePath2,
                                      double tolerance, size_t chunkSize) {
     SF_INFO sfInfo1, sfInfo2;
-    SNDFILE* sndFile1 = sf_open(filePath1.c_str(), SFM_READ, &sfInfo1);
+    SNDFILE* sndFile1 = sf_open(filePath1.string().c_str(), SFM_READ, &sfInfo1);
```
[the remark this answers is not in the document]
omeryusufyagci: In that case, adding a brief comment here would be good as this can easily be forgotten and removed during a refactor. A bit surprising though, as I had no issue running this natively on Windows 10.

launcher.py
```diff
@@ -0,0 +1,405 @@
...
+        input()  # Block until the user presses Enter
+
+    except Exception as e:
+        print(f"An error occurred: {e}")
+        sys.exit(1)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Setup for MediaProcessor Application")
+    parser.add_argument("--mode", choices=["web"], help="Specify mode to launch")
```
omeryusufyagci, line 375: I'd go with `--app` instead of `--mode` here as we want to launch a specific application.
Also, we'd like to make it easy on the user; so if they simply run this launcher, it should install everything and default to the web app if no arguments were provided.

launcher.py
```diff
@@ -0,0 +1,405 @@
...
+        subprocess.CalledProcessError: If any command fails.
+    """
+    if current_debug_level == DebugLevel.ALL:
+        print(f"Executing command: {command}")
+    subprocess.check_call(
+        command.split(),
+        cwd=cwd,
+        stdout=None if current_debug_level == DebugLevel.ALL else subprocess.DEVNULL,
+        stderr=None if current_debug_level == DebugLevel.ALL else subprocess.DEVNULL,
+    )
```
omeryusufyagci, line 47: Would be good to try except here and other places where it's quite possible that we get an exception.

launcher.py
```diff
@@ -0,0 +1,405 @@
...
+        global msys2_root_path
+        installer_url = (
+            "https://github.com/msys2/msys2-installer/releases/download/nightly-x86_64/msys2-base-x86_64-latest.sfx.exe"
+        )
+        installer_name = "msys2-installer.exe"
+
+        msys2_root_path = "C:\\msys64"
+
+        print("Downloading MSYS2 installer...")
+        run_command(f"curl -L -o {installer_name} {installer_url}")
```
omeryusufyagci, line 225: Possible that the host may have connectivity issues.

launcher.py
```diff
@@ -0,0 +1,405 @@
...
+
+        print("Updating MSYS2 packages...")
+        run_command(f"{msys2_root_path}\\usr\\bin\\bash.exe -lc 'pacman -Syu --noconfirm'")
+
+        print("Editing Environment Variables...")
+        # Set it permanently for the current user
+        commands = f"""
+        $oldPath = [Environment]::GetEnvironmentVariable("Path", "User")
+        $newPath = "{msys2_root_path}\\usr\\bin;{msys2_root_path}\\mingw64\\bin;{msys2_root_path}\\mingw32\\bin;" + $oldPath
+        [Environment]::SetEnvironmentVariable("Path", $newPath, "User")
```
omeryusufyagci, line 238: We should clearly log that we changed X for Y **permanently**

launcher.py
```diff
@@ -0,0 +1,405 @@
...
+                if type(cmd) == str:
+                    run_command(cmd)
+                else:
+                    cmd()
+
+        except (FileNotFoundError, subprocess.CalledProcessError):
+            if not install:
+                print(
+                    f"{self.name} is not installed. Use --install to install dependencies automatically or install manually."
+                )
```
omeryusufyagci, line 123: This should be an error level log. Could you use a logger instead of printing? `loguru` is usually pretty nice.

launcher.py
```diff
@@ -0,0 +1,405 @@
...
+        "libsndfile",
+        {"Windows": "mingw-w64-x86_64-libsndfile", "Linux": "libsndfile1-dev"},
+        {"all": ["pkg-config --exists sndfile"]},
+    ),
+    DependencyHandler(
+        "nlohmann-json",
+        {"Windows": "mingw-w64-x86_64-nlohmann-json", "Linux": "nlohmann-json3-dev"},
+        {"all": ["pkg-config --exists nlohmann_json"]},
+    ),
+]
```
omeryusufyagci, line 203: I won't push for this and we can come back to it at a later date, but I'd probably go for something like a `dataclass` here instead of just putting them in the global scope. Same for the webbrowser line with the address and port number. Not critical or urgent, but something to keep in mind

omeryusufyagci (commented):
Thanks for the changes @nikhiljangra264! I've also tested it locally on an Ubuntu system that already had all dependencies installed. Looks good already; please see my comments for details.

MediaProcessor/cmake/src.cmake
nikhiljangra264: It was needed by the command to move the dll to the app's directory. 
Because On windows the mediaprocess.exe doesn't give expected output. So I used dependency walker to check that and find out that the mediaprocessor.exe is unable to find runtime lib.

should i move the the function here?

launcher.py
nikhiljangra264: we can use the argument as
`--mode=app` or `--mode=web`

using --app and --web as arguments have problems like what if a user use both --web and --app, Then both web and app will launch and there will be issue with the backend using same port.

but we can go with --app and --web as seperate arguments if you want.

launcher.py
nikhiljangra264: I purposefully do not handle exception here as exception is handled by the calling entity. This way we can customize exception handling.

launcher.py
nikhiljangra264: I think we should use logging library of the python.
`loguru` will need installation.

launcher.py
omeryusufyagci: Yeah what I meant was like `--app=web` and not `--app` as a standalone arg. Since it's a launcher, we should provide a way to select which app to launch and using the word `mode` feels off to me. 

For the dependency installation and configuration pipeline, we can add something like `--install-dependencies` that would automatically be executed anyway, but would still be nice to provide is to perform an install without launching any app.

launcher.py
omeryusufyagci: Yeah that makes sense, I missed you clearly document that in the doctoring as well. Let's keep it as is

launcher.py
omeryusufyagci: Seeing this again, we should actually check internet access at the start and prompt the user in case of an issue.

MediaProcessor/cmake/src.cmake
omeryusufyagci: Okay makes sense. I probably had the dll on my path and didn't notice; good catch

launcher.py
nikhiljangra264: oh! I get it.

launcher.py
nikhiljangra264: I only able to correct the issues above in available time. I will fix this.

config.json
```diff
@@ -1,8 +1,8 @@
 {
-    "deep_filter_path": "MediaProcessor/res/deep-filter-0.5.6-x86_64-unknown-linux-musl",
-    "deep_filter_tarball_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx.tar.gz",
-    "deep_filter_encoder_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/enc.onnx",
-    "deep_filter_decoder_path": "MediaProcessor/res/DeepFilterNet3_ll_onnx/tmp/export/df_dec.onnx",
+    "deep_filter_path": "MediaProcessor\\res\\deep-filter-0.5.6-x86_64-unknown-linux-musl",
+    "deep_filter_tarball_path": "MediaProcessor\\res\\DeepFilterNet3_ll_onnx.tar.gz",
+    "deep_filter_encoder_path": "MediaProcessor\\res\\DeepFilterNet3_ll_onnx\\tmp\\export\\enc.onnx",
+    "deep_filter_decoder_path": "MediaProcessor\\res\\DeepFilterNet3_ll_onnx\\tmp\\export\\df_dec.onnx",
```
omeryusufyagci, line 5: This may fix it for windows, but breaks it for Unix. While testing this PR on macOS I got some panics from the library which was caused by a file not found due to this. This happened after running the launcher. 

Please dynamically update the paths for Windows and let's keep the default paths with the Unix slashes. There seems to be an issue with the dynamic update of this. Could you please check ? Thanks

nikhiljangra264: The issue is launcher.py successfully updates the path but config.json modified can be accidently pushed with other changes (In this case).

we can change them back after running but what if user runs via command python app.py. That will not work because of paths. So for windows development we have to be careful not to push the config.json.

omeryusufyagci: Well, the launcher will be the official method to launch any application or even for setting up the repository. Of course it's open source, so everyone is free to hack-away at the code, but we don't necessarily have to support use cases that fall outside of the officially supported methods -that's why we provide a launcher. 

I agree with you on your other point though. I'd suggest exploring a runtime override (potentially even a second `runtime_config.json`) instead of directly overwriting the file. The config.json would serve as a base. 

On a second note, we should add a check on app.py to warn the user if it's called from outside the launcher.

MediaProcessor/cmake/test.cmake
```diff
@@ -22,8 +23,19 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(fmt)
 
+# Set platform-specific dynamic library for df
+if(APPLE)
+    set(DF_LIBRARY ${CMAKE_SOURCE_DIR}/lib/libdf.dylib)
```
omeryusufyagci, line 28: we don't yet support x86_64 on apple, which is unfortunate, but rosetta emulator just wouldn't work on my machine... So, we should keep consistent here with the src.cmake in terms of CPU check. Also no need for the comment on l26.

launcher.py
```diff
@@ -0,0 +1,479 @@
...
+def check_internet_connectivity():
+    """For Windows check internet connectivity"""
+    try:
+        logging.debug("Checking internet connectivity... ")
+        # Ping Google's public DNS server
+        run_command("ping -n 1 8.8.8.8")
+        logging.debug("Internet connectivity OK")
+    except subprocess.CalledProcessError:
+        logging.error("No internet connection detected.")
+        sys.exit(1)
```
omeryusufyagci, line 252: this should be cross platform and checked for all platforms. We rely on having libsndfile installed on the machine for instance.

config.json
nikhiljangra264: we should add a runtime json that we dont have to warn user about launching outside the launcher.py
we should standardize to use launcher.py

launcher.py
nikhiljangra264: On linux systems package are not deleted permanently and reside on cache, so checking internet connection may not be important for linux. By the way we can check internet connection by just changing `-n` to `-c`.

omeryusufyagci: Sure, but what about a user that has a missing dependency? We should not assume the dependencies will be available by being cached or another means. In any case we should check for connectivity, and as you suggest, you can use `platform.system()` for example and make it generic.

launcher.py
```diff
@@ -230,144 +382,32 @@ def _get_install_commands_windows(self):
...
-
-        run_command(f"{str(pip_path)} install -r requirements.txt")
-        logging.info("Python dependencies installed.")
-    except Exception as e:
-        logging.error(f"Failed to install Python dependencies: {e}", exc_info=True)
-        sys.exit(1)
+def ensure_MediaProcessor_build(system, re_build=False):
+    """Ensure that the MediaProcessor is build. If rebuild is true, it will rebuild the MediaProcessor."""
+    if not MEDIAPROCESSOR_PATH.exists():
+        os.makedirs(MEDIAPROCESSOR_PATH)
```
omeryusufyagci, line 415: don't like two things here: first we're explicit about the name of the C++ binary, which could change and no reason to couple that with the backend. Second, we're not following naming convention due to the `MediaProcessor`

Please name this to `build_processing_engine()`

launcher.py
```diff
@@ -12,94 +12,149 @@
...
-    },
-    "handlers": {
-        "console": {
-            "class": "logging.StreamHandler",
-            "formatter": "simple",
-            "stream": "ext://sys.stdout",
+LOG_LOCK = threading.Lock()
+
+
+class Config:
```
omeryusufyagci, line 21: From what I see this isn't in itself a Config, but rather something that modifies and manages configurations, be it files or processes. In these cases the name should reflect this. Please rename it to `ConfigManager`.

launcher.py
```diff
@@ -12,94 +12,149 @@
...
+        try:
+            with open(Config.CONFIG_FILE_PATH, "r") as config_file:
+                config = json.load(config_file)
 
-    logging.config.dictConfig(DEFAULT_CONFIG)
+            if system == "Windows":
+                logging.info("Updating config file.")
+                for key, value in config.items():
+                    if type(value) == str:
+                        config[key] = value.replace("/", "\\")
```
omeryusufyagci, line 57: Please encapsulate this into `update_config_for_windows()`, as it makes it difficult to reason about `setup_runtime_config()` seeing a windows specific action without a comment, etc.

launcher.py
```diff
@@ -12,94 +12,149 @@
...
+            return
+        logging.info(f"Generating virtual environment at: {Utils.VENV_DIR_PATH}")
+        Utils.run_command(f"{sys.executable} -m venv {str(Utils.VENV_DIR_PATH)} --system-site-packages")
+        logging.info("Successfully generated Virtual environment.")
+
+    @staticmethod
+    def get_virtualenv_folder() -> Path:
+        """
+        get the path of the virtual environment binaries folder
+        """
```
omeryusufyagci, line 142: please stick to the convention of `directory` instead of the `folder, but in this case, this should be named `get_venv_path()` as you clearly say in the docstring, it returns the path and not a folder.

launcher.py
```diff
@@ -12,94 +12,149 @@
...
+        sys.exit(1)
 
 
 def install_msys2():
+    """
+    Installs MSYS2 and Update Path enviornment variable for windows platform
+    """
     try:
         installer_url = (
             "https://github.com/msys2/msys2-installer/releases/download/nightly-x86_64/msys2-base-x86_64-latest.sfx.exe"
```
omeryusufyagci, line 160: I've been meaning to ask this: any reason why you're using the nightly version? Not familiar with it myself, but don't they have a stable release? At least good to see the switch to `latest` instead of a specific version hash...

launcher.py
```diff
@@ -407,107 +422,122 @@ def ensure_MediaProcessor_build(system, re_build=False):
...
+        ]
+        self.system = system
+        self.URL = "http://127.0.0.1:8080"
+        self.tries = 5
+        self.timeout = 0.5
+        self.setup(log_level, log_file)
+
+    def setup(self, log_level: str, log_file: bool):
+        """
+        Installs the required dependencies and Setup Configuration for the web application.
```
omeryusufyagci, line 471: same thing here, "required" is redundant

launcher.py
```diff
@@ -407,107 +422,122 @@ def ensure_MediaProcessor_build(system, re_build=False):
...
+            # Check if the process is still running
+            if app_process.poll() is not None:
+                error_output = app_process.stderr.read()
+                logging.error(f"Error starting the backend: {error_output}")
+                sys.exit(1)
 
-        # Threads to handle stdout and stderr asynchronously
-        threading.Thread(target=log_stream, args=(app_process.stdout, logging.debug), daemon=True).start()
-        threading.Thread(target=log_stream, args=(app_process.stderr, logging.debug), daemon=True).start()
+            webbrowser.open("http://127.0.0.1:8080")
```
omeryusufyagci: port should be parametric as we may provide it as a config field for the user in the future.

launcher.py
```diff
@@ -407,107 +422,122 @@ def ensure_MediaProcessor_build(system, re_build=False):
...
+}
 
 
 def main():
-    parser = argparse.ArgumentParser(description="Setup for MediaProcessor Application")
-    parser.add_argument("--app", choices=["web", "none"], default="web", help="Specify launch mode")
-    parser.add_argument("--install-dependencies", action="store_true", default=True, help="Install dependencies.")
+    parser = argparse.ArgumentParser(description="Setup for MediaProcessor Application.")
+    parser.add_argument("--app", choices=["web", "none"], default="web", help="Specify launch mode (default=web).")
+    parser.add_argument("--install-only", action="store_true", default=False, help="Install dependencies Only.")
```
omeryusufyagci, line 521: typo: Only

launcher.py
nikhiljangra264: Change name to `get_venv_binaries_directory`?

omeryusufyagci: What I meant was to prefer directory over folder in general, but in this case we're getting a path so we should say this in the method name. So, it'd be `get_venv_path()`.

launcher.py
nikhiljangra264: Should i change `MEDIAPROCESSOR_PATH` to `MEDIAPROCESSOR_BUILD_DIRECTORY` or `PRCOESSING_ENGINE_DIRECTORY`?

omeryusufyagci: > Should i change `MEDIAPROCESSOR_PATH` to `MEDIAPROCESSOR_BUILD_DIRECTORY` or `PRCOESSING_ENGINE_DIRECTORY`?

Let's go with `PROCESSING_ENGINE_PATH`. Could you please check in general for consistency. 
I don't mind calling something a directory, but if the variable is holding literally a path, in that case we should call it a path.

launcher.py
nikhiljangra264: Yeah! I get that point. :+1:
what i meant actually was venv directory is virtual_env and with this function we are getting the directory in which the binaries reside so we should name it get_venv_binaries_directory?

omeryusufyagci: Ah, okay I get it now. In that case I'd still prefer `get_venv_binaries_path()` as it's returning a Path object.

launcher.py
nikhiljangra264: changed this to stable.

omeryusufyagci (approved):
Thanks @nikhiljangra264, looks all good for the updates since the last review!

Have you tested this on Win10, Darwin, and Linux? I don't have access to my HW to do platform test for a few more days. 

Otherwise, this is quite OK for an initial implementation. I'll test it over the weekend, and if all goes well, this should get merged before next week!

nikhiljangra264:
> Thanks @nikhiljangra264, looks all good for the updates since the last review!
> 
> Have you tested this on Win10, Darwin, and Linux? I don't have access to my HW to do platform test for a few more days.
> 
> Otherwise, this is quite OK for an initial implementation. I'll test it over the weekend, and if all goes well, this should get merged before next week!

Tested on linux, Windows 11, deleted everthing and download again working fine.

omeryusufyagci (changes requested):
I tested on Darwin macOS.

The lack of any logging threw me off initially. We should tell the user at least the steps the launcher is taking. Not debug level details, but enough to make them understand things are progressing. I thought the launcher was hanging until I started with debug mode.

Then, I ran into an issue with the .dylib pointing to a local hardcoded path instead of @rpath.
I had seen this the last time around and fixed it, but apparently that was gone with the rebase. Now it's fixed again and I've updated the PR as you'll see.

I'll still take a look at it on Windows, but in the meantime please double check to see we haven't missed anything and take another look at the logging we're providing.

Thanks!

nikhiljangra264:
Checked the code.
Please take a look. contributed after Long time. took some time. :)

omeryusufyagci:
> Checked the code. Please take a look. contributed after Long time. took some time. :)

Thanks, and no worries! 

Trying to understand why I wasn't seeing any logs, I discovered that the logging setup is delegated to the WebApplication class.
In general, control flow seems to depend on WebApplication and for generic things in the launcher we should have common methods. 
I have added FIXMEs in-place please check my last commit here to see more details. 

I would recommend adding dedicated methods to setup/initialize and cleanup (if needed) the launcher itself. That would include setting up the logging, resolving common dependencies, etc. So, we shouldn't treat it as if WebApp will be the only user of the launcher.

Issue #90, opened 2024-12-22
Bug: Audio file treated as a video file


Rushi1109:
**Describe the bug**
When uploading a .mp3 file. The backend serves the .mp4 file which is very short in duration. So, The correct output is not being served from backend.

**To Reproduce**
1. Upload or add .mp3 file to server. (which contains cover image)
2. Play the audio when playback widget is available
3. See error

**Expected behavior**
As a user I expect the correct output of my audio file. ( Instead of .mp4 file, The backend should serve correct audio file.)

**Logs and Debug Information**
Please include any relevant log messages and error outputs here.

**Screenshots**

**System and software (please complete the following information):**
 - OS: Windows 10 Intel i5 9300HF
 - Output of `ffmpeg -version` : ffmpeg version 5.1.6-0+deb12u1 Copyright (c) 2000-2024 the FFmpeg developers
built with gcc 12 (Debian 12.2.0-14)

**Additional context**
Please include system settings, configuration changes, or anything else you believe might contribute to the problem.

Rushi1109:
My proposed solution is to maintain an array of audio file extensions. When a file is uploaded, we can check if its extension matches any entry in our array. If a match is found, the file can be treated as an audio file.

omeryusufyagci:
Hi @Rushi1109, thanks for reporting this bug and also the proposed solution! 

For completeness, this was initially reported over at discord, and with your help we understood that the issue likely stemmed from the audio file having a cover picture. This cover image was being mistakenly identified as a very short video stream by `ffprobe`, which is the tool we're using to determine media type to invoke the appropriate pipeline. 

Currently, for an audio file, we do not expect any video streams and the existence of that stream is causing the video pipeline to be erroneously invoked.

Briefly checking this, it seems it's intended behavior for `ffprobe` to list this image as a **video** stream due to the `mjpeg` codec. So, it's not really a bug on ffmpeg's side, but us not realizing this was a possibility yet. 

Regarding how we should fix this, your suggestion of keeping a list of audio file extensions may be a possibility, but I feel it'd be error prone to solely rely on the extension. Given the fact, that `ffprobe` provides the entire list of streams we could deduce if the video streams for a given file should be neglected. 

Revisiting the image you'd provided in discord, below:
![image](https://github.com/user-attachments/assets/827ab452-d1f3-4569-b80c-0889e700fcf6)
We see that the cover picture is reported without a duration and has `mjpeg` codec. 

[...]

PR #91, opened 2024-12-23
Bug: Fix audio file with cover image treated as video file 

Rushi1109:
fixes the bug raised in #90 

The `ffprobe` tool is used to check if any stream contains codec_name as mjpeg.

If mjpeg is present, the media is treated as Audio, other media is treated as Video

### DIFF
--- a/MediaProcessor/src/Engine.cpp
+++ b/MediaProcessor/src/Engine.cpp
@@ -86,18 +86,26 @@ MediaType Engine::getMediaType() const {
 
 bool Engine::hasValidVideoStream(const nlohmann::json& streamData) const {
     for (const auto& stream : streamData["streams"]) {
-        if (stream["codec_type"] != "video") {
-            continue;
-        }
-        if(!stream.contains("avg_frame_rate")) {
-            continue;
-        }
-
-        if (!hasZeroFrameRate(stream["avg_frame_rate"])) {
+        if(isValidVideoStream(stream)) {
             return true;
         }
     }
     return false;
+}
+
+bool Engine::isValidVideoStream(const nlohmann::json& stream) const {
+    if (stream["codec_type"] != "video") {
+        return false;
+    }
+    if(!stream.contains("avg_frame_rate")) {
+        return false;
+    }
+
+    if (hasZeroFrameRate(stream["avg_frame_rate"])) {
+        return false;
+    }
+
+    return true;
 }
 
 bool Engine::hasValidAudioStream(const nlohmann::json& streamData) const {
--- a/MediaProcessor/src/Engine.h
+++ b/MediaProcessor/src/Engine.h
@@ -52,18 +52,17 @@ class Engine {
     MediaType getMediaType() const;
 
     /**
-     * @brief Checks if the streams contains valid video stream and not static media (i.e. cover image)
-     * 
-     * @param streamData json data of all the streams
+     * @brief Checks if the streams contains a valid video stream.
+     * A valid video stream is one that has an actual duration and is not static media (e.g., a cover image).
      *
      * @return True if the stream is valid video stream. False otherwise.
      */
     bool hasValidVideoStream(const nlohmann::json& streamData) const;
 
+    bool isValidVideoStream(const nlohmann::json& stream) const;
+
     /**
-     * @brief Checks if the streams contains valid audio stream
-     * 
-     * @param streamData json data of all the streams
+     * @brief Checks if the streams contains a valid audio stream
      *
      * @return True if the stream is valid. False otherwise.
      */
@@ -71,8 +70,6 @@ class Engine {
 
     /**
      * @brief Checks if the stream's avg_frame_rate is 0/0
-     * 
-     * @param frameRate the avg_frame_rate field from stream
      *
      * @return True if avg_frame_rate equals to 0/0. False otherwise.
      */


omeryusufyagci:
Hey @Rushi1109, thanks for the PR. Frankly, I wasn't familiar with the full output set of `ffprobe`, but from what I see they provide `disposition.attached_pic` as a field in itself. Could you please get the json output which should give you a list of streams, with each stream providing you this field. From what I briefly checked, that should literally be what we're looking for. 

Thanks for checking!

Rushi1109:
Hey @omeryusufyagci,

When looking at the raw data of the media I see `DISPOSITION:attached_pic = 1` for stream index `2` which is the video stream.

But, When I tried to get the information about `disposition.attached_pic` from `ffprobe` tool, I am not able to get it's value. I tried extracting from json but still it's not showing attached_pic for that file. Do you have any idea about why it's getting filtered out, by ffmpeg?

Rushi1109:
Moreover @omeryusufyagci,

I have researched a bit and got to know that If the video stream contains a static image, the average frame rate for that stream will be `0/0`. My current implementation is using the avg_frame_rate from video stream and decides the media type.

Do you see any issue with this approach?

omeryusufyagci:
> Hey @omeryusufyagci,
> 
> When looking at the raw data of the media I see `DISPOSITION:attached_pic = 1` for stream index `2` which is the video stream.
> 
> But, When I tried to get the information about `disposition.attached_pic` from `ffprobe` tool, I am not able to get it's value. I tried extracting from json but still it's not showing attached_pic for that file. Do you have any idea about why it's getting filtered out, by ffmpeg?

Please make sure you include `disposition` after the `-show_entries` flag, e.g. `-show_entries stream=index,<other_stuff>,disposition`. 

Could you try like that, and if there's still an issue please provide the command you used and the output you received. 

> I have researched a bit and got to know that If the video stream contains a static image, the average frame rate for that stream will be 0/0. My current implementation is using the avg_frame_rate from video stream and decides the media type.

> Do you see any issue with this approach?

Thanks for figuring this out! Good to know that'd be a valid fix. The `disposition.attached_image` would be a more robust solution, as it explicitly identifies cover images and aligns better with the intent of filtering out such streams.

Could you please verify this and let me know if it resolves the issue? Thank you!

Rushi1109:
> Please make sure you include `disposition` after the `-show_entries` flag, e.g. `-show_entries stream=index,<other_stuff>,disposition`

Yes I included `disposition`. I am using the same command of `ffprobe` but instead of avg_frame_rate, I added disposition and removed that awk command. But, I am not seeing any output for disposition.

Can you check it on your side? Do you see disposition in output?

omeryusufyagci:
> > Please make sure you include `disposition` after the `-show_entries` flag, e.g. `-show_entries stream=index,<other_stuff>,disposition`
> 
> Yes I included `disposition`. I am using the same command of `ffprobe` but instead of avg_frame_rate, I added disposition and removed that awk command. But, I am not seeing any output for disposition.
> 
> Can you check it on your side? Do you see disposition in output?

HI @Rushi1109, you're right. I was able to reproduce the same behavior. It depends on what's bundled with the metadata, and we won't be able to count on the existence of that field. That means we'll have to check for this ourselves. 

You're also correct, that a cover image is reported as a video stream that has `0/0` avg frame rate. It may also be reported as 'Cover' in a `comment` field, but that again is dependent on the metadata, whereas the 0/0 avg frame rate seems to be always available. 

In this case, I propose to get the same ffprobe output as json (fyi, we're already using nlohmann/json in the project), parse the avg frame rate into a vector, and check if a given file has only 1 video stream **and** that video stream has 0/0 avg frame rate. Meaning, as soon as we iterate through a second video stream, we should conclude our check to avoid wasting time there. 

Could you update this PR to implement this? Thanks

Rushi1109:
> HI @Rushi1109, you're right. I was able to reproduce the same behavior. It depends on what's bundled with the metadata, and we won't be able to count on the existence of that field. That means we'll have to check for this ourselves.
> 
> You're also correct, that a cover image is reported as a video stream that has `0/0` avg frame rate. It may also be reported as 'Cover' in a `comment` field, but that again is dependent on the metadata, whereas the 0/0 avg frame rate seems to be always available.
> 
> In this case, I propose to get the same ffprobe output as json (fyi, we're already using nlohmann/json in the project), parse the avg frame rate into a vector, and check if a given file has only 1 video stream **and** that video stream has 0/0 avg frame rate. Meaning, as soon as we iterate through a second video stream, we should conclude our check to avoid wasting time there.
> 
> Could you update this PR to implement this? Thanks

Sure. I'll look into that today.

Rushi1109:
Hello @omeryusufyagci, 

I have made required changes. I have tested it manually and it is working as expected. 

Can you test it on your side and let me know if any change is required?

omeryusufyagci (changes requested):
Thanks for the changes @Rushi1109 !

Here's some comments for you:

**About the `parseFrameRate()` addition**:
The `MediaProcessor::Utils` NS is normally meant for generic utility. In this case, the frame rate helper is intrinsically coupled with `getMediaType()`. So, we should move this under the Engine class as a private member function. Also, for expressiveness, let's rename it to `hasZeroFrameRate()`.

Regarding the implementation of this helper, I believe we can make it simpler. We want to know if an otherwise audio file has a video stream that has exactly "0/0" for the avg frame rate. So, we can simply do a full string comparison, instead of parsing each char in the field. Pass a const ref of the stream object, and it can be as simple as a 1-line return statement that checks if it matches the expected case. We should however, make sure to cover the cases where the field is missing or invalid. 

**Updates on `getMediaType()`**:
Good start on this with the json parsing, but we can also simplify this similar to above. We should also encapsulate the logic where we're determining if we have a valid video stream. Otherwise, it makes it needlessly difficult understand what this function does at a glance. 

Here's roughly what I have in mind:

```cpp
for (const auto& stream : streams) {
    if (hasValidVideoStream(stream)) {
        return MediaType::Video;
    }
    if (hasValidAudioStream(stream)) {
        hasAudioStream = true; // define this flag false before the loop
    }
}

// if we haven't returned for Video and have found a valid audio stream
if (hasAudioStream) {
    return MediaType::Audio;
}

// otherwise
return MediaType::Unsupported;
}
```

So please encapsulate both checks (`hasValid...`) into individual functions. I realize the one for audio will likely be a 1-line return statement, but I'd still prefer to have it for the sake of consistency. 
Within the `hasValidVideoStream()`, you'd call `hasZeroFrameRate()` with a const ref of the json stream for the avg frame rate field.
The other 2 helpers should also be called with const refs of the stream and all 3 new functions should be const themselves too. 

This approach should make it easier to understand what `getMediaType()` does, and in general simplify the deduction logic to see if we have a cover image. Could you please update this PR accordingly?

Please let me know if you have any questions along the way. Many thanks!

Rushi1109:
Updated the PR.

MediaProcessor/src/Engine.h
```diff
@@ -26,6 +27,8 @@ class Engine {
 
    private:
     std::filesystem::path m_mediaPath;
+    mutable bool hasAudioStream;
+    mutable int videoStreamCount;
```
omeryusufyagci: The reason to add these as member variables is not clear to me. They do not represent persistent state for the Engine class. Their meanings persist through a function call of `getMediaType()` and do not really make sense outside of that. 

Although not critical here as we won't keep them, but for the future: before adding mutable-specified variables, we should take a good look at why we intend to that. It could point to a bad design choice that I'd rather address at source than to go the `mutable` route.

Please refactor these variables to be local to the respective function, where they're used.

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +67,49 @@ bool Engine::processVideo() {
...
+    // If we haven't returned for Video and have found a valid audio stream
+    if (hasAudioStream) {
         return MediaType::Audio;
-    } else {
-        throw std::runtime_error("Unsupported media type detected.");
     }
+
+    // otherwise
+    return MediaType::Unsupported;
+}
```
omeryusufyagci: The 2 lines of comments here were more for providing clarity in the discussion, but typically we refrain to add comments that state otherwise obvious operations. So in this case, let's remove the comments here

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +67,49 @@ bool Engine::processVideo() {
...
-    } else if (result.find("audio") != std::string_view::npos) {
+    nlohmann::json streamData = nlohmann::json::parse(*output);
+
+    for (const auto& stream : streamData["streams"]) {
+        if (hasValidVideoStream(stream)) {
+            return MediaType::Video;
+        } else if (hasValidAudioStream(stream)) {
+            hasAudioStream = true;
+        }
+    }
```
omeryusufyagci: l.82 here the else if is redundant as we're returning for valid video above. It should be a separate if statement

MediaProcessor/src/Engine.h
```diff
@@ -49,6 +52,33 @@ class Engine {
...
     MediaType getMediaType() const;
+
+    /**
+     * @brief Checks if the provided stream is valid video stream
+     * 
+     * @param stream json data of stream
+     *
+     * @return True if the stream is valid. False otherwise.
+     */
+    bool hasValidVideoStream(const nlohmann::json& stream) const;
```
omeryusufyagci: What does _valid_ mean here? We should describe what exactly that means here

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +67,49 @@ bool Engine::processVideo() {
...
-        throw std::runtime_error("Unsupported media type detected.");
     }
+
+    // otherwise
+    return MediaType::Unsupported;
+}
+
+bool Engine::hasValidVideoStream(const nlohmann::json& stream) const {
+    if (stream["codec_type"] == "video") {
+        ++videoStreamCount;
```
omeryusufyagci: Incrementing this here (even after the refactor to make it a local var to getMediaType) would make it more difficult to understand the workflow. This should be done at the call side.

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +67,49 @@ bool Engine::processVideo() {
...
-    } else {
-        throw std::runtime_error("Unsupported media type detected.");
     }
+
+    // otherwise
+    return MediaType::Unsupported;
+}
+
+bool Engine::hasValidVideoStream(const nlohmann::json& stream) const {
+    if (stream["codec_type"] == "video") {
```
omeryusufyagci: Also adding to my comment below, this should be an early return. If it's not a video stream to begin with, we should immediately return false.

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +67,49 @@ bool Engine::processVideo() {
...
+        ++videoStreamCount;
+
+        if (videoStreamCount > 1 || !hasZeroFrameRate(stream["avg_frame_rate"])) {
+            return true;
+        }
```
omeryusufyagci: This does not cover the inexistence of the field, and may cause a crash. Please ensure the field is available and return false otherwise. Similarly, do a return false, instead of nesting under a conditional level

omeryusufyagci (changes requested):
Hi @Rushi1109, thanks for the updates. Definitely moving in the right direction! Could you please take a look and update the PR? Thanks!

MediaProcessor/src/Engine.cpp
Rushi1109: > l.82 here the else if is redundant as we're returning for valid video above. It should be a separate if statement

How is it redundant? I don't understand. Can you please explain it?

MediaProcessor/src/Engine.cpp
Rushi1109: > Incrementing this here (even after the refactor to make it a local var to getMediaType) would make it more difficult to understand the workflow. This should be done at the call side.

Removing this variable. As we will return from `getMediaType()` as soon as we will get an actual video stream.(Video stream that is not cover image.)

In case when we encounter video stream with cover image, we will return false from here. In next iteration, when an actual video stream is computed, we will return true after checking `avg_frame_rate` of that stream. I do suggest to check for `avg_frame_rate` for that stream as well. 

I don't think we should return just by video stream count being more than 1. We should check for avg_frame_rate. What do you think about this? let me know.

omeryusufyagci: I've checked the changes. 

First off, we're now saying any stream that has the avg frame rate field will be treated as an invalid video stream. What if ffprobe returns sensible values for actual videos with that field? which wouldn't be surprising
Checking if the frame rate is exactly 0/0 and invalidating then, is logical. To determine if a video stream is valid, perhaps you can check non-zero duration in addition.

Also, I don't see the point of refactoring the function name to `isActualVideoStream`.

Regarding your question: As soon as we have even one single valid video stream, that would classify the media file as a video and we should return Video. There is no longer a need to check each of the remaining streams. 
On the contrary, to determine if an audio file only has a cover attached, we have to ensure that media file has only 1 video stream whose avg frame rate is exactly 0/0. 

I hope this answers your question, and please let me know otherwise.

MediaProcessor/src/Engine.cpp
omeryusufyagci: The two conditional checks have no logical coupling or dependency. Since we already return if a valid video stream is found, the code following that `if` statement will only execute if the condition was false. So, the `else if` is redundant, as the second check is independent and can simply be a separate if.

MediaProcessor/src/Engine.cpp
Rushi1109: > First off, we're now saying any stream that has the avg frame rate field will be treated as an invalid video stream. What if ffprobe returns sensible values for actual videos with that field? which wouldn't be surprising
> Checking if the frame rate is exactly 0/0 and invalidating then, is logical. To determine if a video stream is valid, perhaps you can check non-zero duration in addition.

The current implementation works as described below:
1) It checks for the `codec_type` field. It should be video.
2) It then ensures the `avg_frame_rate` field exists.
3) And then we check if the `avg_frame_rate` is equal to `0/0` which indicates it is a static media.

In all the above cases we return false.

4) Otherwise we return true as the stream is now actual video stream.

The thing with the duration field would be problematic, because when we checked earlier, the ffmpeg tool outputs the duration of static media same as audio duration. So, I don't think we should rely on duration field.

The current implementation look good to me. As there is early return when we encounter first ever valid video stream. (We terminate the for loop from that point)

But, we do iterate when the first stream is not valid video stream (static media, cover image).

Let me know if there's something I've misunderstood.

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +65,52 @@ bool Engine::processVideo() {
...
+    }
+
+    if (containsAudioStream) {
         return MediaType::Audio;
-    } else {
-        throw std::runtime_error("Unsupported media type detected.");
     }
+
+    return MediaType::Unsupported;
+}
```
omeryusufyagci: Since the helpers are called hasValid...Stream it could be nicer to pass the entire stream and handle the iterations within the helpers. In doing so, I feel there won't be a need for the audio helper flag `containsAudioStream`.
You can check if we have a valid video stream and return Video, otherwise check for audio and return Audio, otherwise return Unsupported. 

This would make getMediaType very easy to reason about.

MediaProcessor/src/Engine.cpp
omeryusufyagci: Hi @Rushi1109, sorry for the delay on this. From what I see in the commit history these changes were introduced after my comment above, which is great, but I was unsure for a second when you said _the current implementation_. 

Anyway, that's not really important. Please see my comment below to finalize this PR. I think it's good enough to merge and will definitely be useful!

MediaProcessor/src/Engine.cpp
Rushi1109: > Since the helpers are called hasValid...Stream it could be nicer to pass the entire stream and handle the iterations within the helpers. In doing so, I feel there won't be a need for the audio helper flag `containsAudioStream`.
> You can check if we have a valid video stream and return Video, otherwise check for audio and return Audio, otherwise return Unsupported.

Done. I have made the changes. Can you please review it?

MediaProcessor/src/Engine.cpp
```diff
@@ -65,23 +65,52 @@ bool Engine::processVideo() {
...
+
+bool Engine::hasValidVideoStream(const nlohmann::json& streamData) const {
+    for (const auto& stream : streamData["streams"]) {
+        if (stream["codec_type"] != "video") {
+            continue;
+        }
+        if(!stream.contains("avg_frame_rate")) {
+            continue;
+        }
```
omeryusufyagci: It would be better to check for the inverse condition and return false, instead of continuing for the expected case. The earlier implementation in getMediaType was like that already.

MediaProcessor/src/Engine.h
```diff
@@ -49,6 +50,33 @@ class Engine {
...
+    bool hasValidAudioStream(const nlohmann::json& streamData) const;
+
+    /**
+     * @brief Checks if the stream's avg_frame_rate is 0/0
+     * 
+     * @param frameRate the avg_frame_rate field from stream
+     *
+     * @return True if avg_frame_rate equals to 0/0. False otherwise.
+     */
+    bool hasZeroFrameRate(const std::string& frameRate) const;
```
omeryusufyagci: So far we have not included  `[at]param`s for function parameters in the project. Let's keep it that way, and eventually we'll likely do a review of our commenting practices in the public APIs. Anyway, for private member functions, it's not necessary to provide docs in the header file. 

Also, my previous remark on what valid stands for, for the 2 helpers remain open. I would add a line in the description, i.e. after the `brief` to define what valid means in our context.

omeryusufyagci (approved):

Issue #56, opened 2024-10-24
Improvement: Add defaults and settings validation to `FFmpegSettingsManager`


omeryusufyagci:
Currently the FFmpegSettingsManager provides an interface for the FFmpeg user settings.
However, it should also provide a set of sensible defaults for ease of use, and it validate each set operation performed on it.

andcscott:
Hey @omeryusufyagci, wanted to check if the revisions for #60 (commit 467fe76ea57938ad2509c62bb05f4220eb6e4b85 on my branch) took care of this? If any refactoring is needed I'll have some time this week/next week if you'd like.

omeryusufyagci:
Hey @andcscott, thanks for checking in! Yes, it has addressed it partially. Now we have defaults which is great, but I also want to ensure the config manager is entirely responsible for the validation of the fields as well. There are cases where some codecs won't work with certain other settings and of course the basic validation of each field to check if they're within bounds, and are logical to a point, etc.

If you'd like to take a pass at it, I'd gladly accept the help ! : )

PR #95, opened 2025-01-27
Feature: Add settings validation to `FFmpegConfigManager`

andcscott:
Hi @omeryusufyagci, hope you're doing well!

This PR relates to #56 by adding settings validation for the current FFmpeg options available in `FFmpegConfigManager`. More specifically, it attempts to address issues where the sample rate or number of audio channels could be unacceptable to FFmpeg. I reviewed the codecs rather thoroughly, but I did not consider other audio/video options for this PR primarily because they aren't currently needed/I'm unsure which you'd eventually like to add.

In addition to referencing the current specs for the available codecs (AAC, MP3, FLAC, Opus) I also did some testing with FFmpeg to determine when it continue even when given an option outside the spec. In this way, the config manager should only throw an error when FFmpeg would truly panic, not when it proceeds with its own set of assumptions. I also reviewed cases where FFmpeg could throw an error even though the spec states it is valid (255 Opus channels could require a very complicated rematrix!).

Last, two template functions were added to help streamline the process of adding new options later. These templates address the two primary ways in which FFmpeg conceives of a valid option: either a maximum unsigned integer, or a list of acceptable unsigned integers.

### DIFF
--- a/MediaProcessor/src/FFmpegConfigManager.cpp
+++ b/MediaProcessor/src/FFmpegConfigManager.cpp
@@ -1,7 +1,5 @@
 #include "FFmpegConfigManager.h"
 
-#include <unistd.h>
-
 #include <filesystem>
 #include <iostream>
 #include <stdexcept>
@@ -36,14 +34,11 @@ void FFmpegConfigManager::setCodecStrictness(CodecStrictness strictness) {
 }
 
 void FFmpegConfigManager::setInputFilePath(const fs::path inputFilePath) {
-    int read = access((inputFilePath.string().c_str()), R_OK);
-    if (read < 0 && errno == EACCES) {
-        throw std::runtime_error("Insufficient permissions to read input file!");
-    } else if (read < 0) {
-        throw std::runtime_error("Unable to read input file, are you sure it exists?");
-    } else {
-        m_globalSettings.inputFilePath = inputFilePath;
+    fs::perms p = fs::status(inputFilePath).permissions();
+    if ((p & fs::perms::owner_read) != fs::perms::owner_read) {
+        throw std::runtime_error("Error: Permission denied when reading input file");
     }
+    m_globalSettings.inputFilePath = inputFilePath;
 }
 
 void FFmpegConfigManager::setOutputFilePath(const fs::path outputFilePath) {
--- a/MediaProcessor/src/FFmpegConfigManager.cpp
+++ b/MediaProcessor/src/FFmpegConfigManager.cpp
@@ -130,51 +130,63 @@ void FFmpegConfigManager::updateSettings(const struct FFmpegGlobalSettings& glob
 
 // Validate Settings
 template <typename T>
-void FFmpegConfigManager::validateOption(const T optionValue, const T optionMax) {
-    if (optionValue > optionMax) {
+void FFmpegConfigManager::validateOption(const T value, const T min, const T max) {
+    if (((value - min) | (max - value)) < 0) {
         std::string codec =
             Utils::enumToString<AudioCodec>(getAudioCodec(), getAudioCodecAsString());
-        throw std::runtime_error("Invalid Audio Option \"" + std::to_string(optionValue) +
-                                 "\" : FFmpeg accepts a maximum of " + std::to_string(optionMax) +
-                                 " for " + codec);
+        throw std::runtime_error("Invalid Audio Option \"" + std::to_string(value) +
+                                 "\" : FFmpeg accepts values between " + std::to_string(min) +
+                                 " and " + std::to_string(max) + " for " + codec);
     }
 }
 
-template <typename T, std::size_t N>
-void FFmpegConfigManager::validateOption(const T optionValue,
-                                         const std::array<T, N>& validOptions) {
-    for (const T& valid : validOptions) {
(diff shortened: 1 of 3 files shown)


MediaProcessor/src/FFmpegConfigManager.cpp
```diff
@@ -32,10 +36,18 @@ void FFmpegConfigManager::setCodecStrictness(CodecStrictness strictness) {
...
-    m_globalSettings.inputFilePath = inputFilePath;
+    int read = access((inputFilePath.string().c_str()), R_OK);
+    if (read < 0 && errno == EACCES) {
+        throw std::runtime_error("Insufficient permissions to read input file!");
+    } else if (read < 0) {
+        throw std::runtime_error("Unable to read input file, are you sure it exists?");
+    } else {
+        m_globalSettings.inputFilePath = inputFilePath;
+    }
 }
```
omeryusufyagci: Can you use the file system API here instead? [this](https://en.cppreference.com/w/cpp/filesystem/permissions) should have all you'd need to do the check.

MediaProcessor/src/FFmpegConfigManager.cpp
```diff
@@ -84,6 +96,10 @@ int FFmpegConfigManager::getAudioChannels() const {
 // Video Setters
 void FFmpegConfigManager::setVideoCodec(VideoCodec codec) {
     m_videoSettings.codec = codec;
+    if (codec != VideoCodec::COPY) {
+        std::cerr << "Warning: Consider setting video codec to COPY if quality or conversion time are "
+                "unacceptable" << std::endl;
+    }
```
omeryusufyagci: This took me a second to get it, so perhaps consider something like "Consider setting the video codec to COPY for higher quality and faster conversion times."

MediaProcessor/src/FFmpegConfigManager.cpp
```diff
@@ -116,4 +132,63 @@ void FFmpegConfigManager::updateSettings(const struct FFmpegGlobalSettings& glob
...
+            setAudioChannels(2);
+            setAudioSampleRate(48000);
+            std::cerr << "Warning: Audio codec cannot be UNKNOWN, default to AAC @ 48000 Hz with 2 "
+                         "channels"
+                      << std::endl;
+            break;
+        default:
+            return;
+    }
+}
```
omeryusufyagci: Quite okay to use switch case as we don't expect this list to grow (not too much at least), but mostly to reduce verbosity and make it more in line with the rest of the core, please consider implementing this with an unordered map, e.g. codecValidationRules that would map each codec to a struct that would hold the data needed for your validation (like max channels and sample rate). 

The benefit would then be to simply try to find it in the map and handle the case it isn't found. This would also add some flexibility for when we might need to add rules for codec validation.

MediaProcessor/src/FFmpegConfigManager.cpp
```diff
@@ -116,4 +132,63 @@ void FFmpegConfigManager::updateSettings(const struct FFmpegGlobalSettings& glob
...
+        if (m_audioSettings.sampleRate == valid) return;
+    }
+    std::string codec = Utils::enumToString<AudioCodec>(getAudioCodec(), getAudioCodecAsString());
+    std::ostringstream ssError;
+    ssError << "Invalid " + codec + " sample rate: FFmpeg supports ";
+    for (const T& opt : validOptions) {
+        ssError << opt << " ";
+    }
+    throw std::runtime_error(ssError.str());
+}
```
omeryusufyagci: I would go with vector here for validOptions as it would make it a bit more flexible and keep the template parameters cleaner. Vector should be almost as good, with such a small data set and likely infrequent access patterns.

omeryusufyagci (changes requested):
Hi @andcscott, thanks a lot for submitting this! 

I'd say it's in a pretty good shape already. Only _serious_ thing we should take a look at is the `setInputFilePath` function, that I commented on. The rest are more for general styling and usage uniformity across the core. Please take a look and let me know if you have any feedback! 

Should be good to go following these changes. Thanks again !

andcscott:
Hey @omeryusufyagci, while I'm not super happy with how this turned out, it's hardly generic, it will hopefully help progress a little.

[18ecf49](https://github.com/omeryusufyagci/fast-music-remover/pull/95/commits/18ecf49515251d8f330abcb3e36a370aec29e59f): pretty simple here, just replaced the C-style permissions check with the filesystem API

[68fbab5](https://github.com/omeryusufyagci/fast-music-remover/pull/95/commits/68fbab57da0e49ac606675c9b9a076bf0cecda02): another simple one, reworded the warning as requested

[dc7ec59](https://github.com/omeryusufyagci/fast-music-remover/pull/95/commits/dc7ec595b0080749db16bfacb813544ccc6a5e8b): while tempted to go all-in with the unordered map I'm glad I checked with you on Discord, hopefully this will suffice for now

[51c57a8](https://github.com/omeryusufyagci/fast-music-remover/pull/95/commits/51c57a8bbf5caeaf8896e8f1598b3996434241a9): like the above, it feels more "proper" to use an interface/strategy, but the benefits just weren't there yet, let me know if you've changed your mind on that or see a better way

Thanks!

